PR #2, opened 2018-05-14
Update README.md

jdb:
Hi Flavio, can I send the pull request to ariava?
I tried but 'ariava' but the UI says 'Nothing to show'. I guess it does not matter too much for codeu.


flerdacodeu (approved):

flerdacodeu:
You can only assign reviews to collaborators. I invited to each group repository only the mentor for that group because otherwise it would be confusing to people (each mentor would have 11 group repositories). But you can add them as a collaborator and then send them a pull request for review.

PR #3, opened 2018-05-20
Sridha2412 1

Sridha2412:


sridha2412/assignment1/string_anagrams.py
```diff
@@ -0,0 +1,48 @@
...
+# assumption: only using the 26 alphabets of the english language
+# method to check if two strings are anagrams
+# method takes two strings and a boolean value. If true, check case sensitive anagrams else ignore case
+def isAnagram(str1, str2, caseSensitive):
+    # if ignoring case, convert both strings to lower case
+    if not caseSensitive:
+        str1 = str1.lower()
+        str2 = str2.lower()
+    # sort and compare the strings
+    return sorted(str1) == sorted(str2)
```
jdb: I see a duplication of the logic: the same processing is done on str1 and str2, and then below on each of the words of each of the sentence. Do you think you can propose a solution to encapsulate this logic to be able to reuse it?

sridha2412/assignment1/string_anagrams.py
```diff
@@ -0,0 +1,48 @@
...
+
+# assumption: sentence can only include alphabets and spaces
+# method to check if two sentences are anagrams
+# method takes two sentences and a boolean value. If true, check case sensitive anagrams else ignore case
+def isSentenceAnagram(str1, str2, caseSensitive):
```
jdb: I think a required condition for two sentences to be anagrams is for each of the word in sentence1 to have an anagram word in sentence2. Do you think you can re-use isAnagram() in IsSentenceAnagram?

sridha2412/assignment1/string_anagrams.py
```diff
@@ -0,0 +1,48 @@
...
+    # sort and compare the strings
+    return sorted(str1) == sorted(str2)
+
+str1 = input("Enter first string: ")
+str2 = input("Enter second string: ")
+caseSensitive = input("Enter Y if case sensitive else N: ")
+# default value is False
+caseSensitive = True if caseSensitive == 'Y' else False
+
+# for words
```
jdb: Please remove the "dead" code

sridha2412/assignment1/string_anagrams.py
```diff
@@ -0,0 +1,48 @@
+# python program to determine if two strings are anagrams of each other
+
+# assumption: only using the 26 alphabets of the english language
+# method to check if two strings are anagrams
+# method takes two strings and a boolean value. If true, check case sensitive anagrams else ignore case
+def isAnagram(str1, str2, caseSensitive):
```
jdb: Please check out https://google.github.io/styleguide/pyguide.html#Naming

Python usually do not use a lot the "camel case" convention but instead uses lower case and underscore.

sridha2412/assignment1/string_anagrams.py
```diff
@@ -0,0 +1,48 @@
...
+str1 = input("Enter first string: ")
+str2 = input("Enter second string: ")
+caseSensitive = input("Enter Y if case sensitive else N: ")
+# default value is False
+caseSensitive = True if caseSensitive == 'Y' else False
+
+# for words
+# if isAnagram(str1, str2, caseSensitive):
+# for sentences
+if isSentenceAnagram(str1, str2, caseSensitive):
```
jdb: Think about using the ternary operator, you would get less duplication
is_it_or_not = '' if isSentenceAnagram(str1, str2, caseSensitive) else 'NOT '
print(...)

sridha2412/assignment1/string_anagrams.py
```diff
@@ -0,0 +1,48 @@
...
+    print("\""+str1+"\" and \""+str2+"\" are anagrams")
```
jdb: No need to the slash: you can alternate the double and single quote: this is quite convenient.

Ex:
print("I'm embedding a single quote without slashes.")

print('What "they" want us to believe is that "we" "landed" on "the moon"')
print('Wake up, ' + "Sheeple !")

jdb (changes requested):
Very cool, my first CodeU pull request, good job you :)

sridha2412/assignment1/string_anagrams.py
Sridha2412: I have made all necessary changes. However, I am not sure sentence_anagram(..) is very efficient.

sridha2412/assignment1/string_anagrams.py
```diff
@@ -14,35 +14,31 @@ def isAnagram(str1, str2, caseSensitive):
...
-    for i, word in enumerate(str2):
-        str2[i] = sorted(list(word))
-
-    # sort and compare the strings
-    return sorted(str1) == sorted(str2)
+    # check both have equal number of words
+    if len(str1) != len(str2):
+        return False
+    # iterate over both lists
+    for i in range(0, len(str1)):
```
jdb: 2 notes: 
- range(0, 10) is equivalent to range(10). When the first element is not set, then its value defaults to zero.
- this is a problematic change: the previous algorithm was O(n * ln (n)) but now it is O(n * m) where n is the number of words in the first sentences and is the number of words in the second sentence.

I suggest you create a function that takes a sentence and returns a structure whose equality method test for anagrams. This function can take the sentence as an input and return a list of occurences count dictionaries. Do not forget to sort the words of the sentence before creating the list.

I will send a solution of anagram today.

jdb (changes requested):
Ah, there is an introduction of O(n^2) algorithm over the previous version that was O(n * ln (n)). This is a regression :)

jdb (approved):

PR #5, opened 2018-05-21
 randomUsernameAndPassword - first assignment

randomUsernameAndPassword:


randomUsernameAndPassword/assignment1/Q1.py
```diff
@@ -0,0 +1,31 @@
...
+
+def is_anagram(str1, str2, case_sensitive):
+    
+    if not case_sensitive:
+        str1.lower()
+        str2.lower()
+    
+    str1 = str1.split(" ")
+    str2 = str2.split(" ")
+    str1 = sorted([''.join(sorted(i)) for i in str1])
```
jdb: Short and powerful: very cool/

In this code you duplicate the logic to "normalize" a sentence: do you want to create a function that takes the sentence and return a structure that is comparable?

randomUsernameAndPassword/assignment1/Q2.py
```diff
@@ -0,0 +1,54 @@
...
+  def __init__(self, values):
+    assert(values)
+    head, tail = values[0], values[1:]
+    self.data = head
+    if tail:
+      self.next = Node(tail)
+    else:
+      self.next = None
+
+  def get_next(self):
```
jdb: Please use direct access to attribute, this is common in Python
https://google.github.io/styleguide/pyguide.html?showone=Properties#Properties

What advantages do you see with the getters?

randomUsernameAndPassword/assignment1/Q2.py
```diff
@@ -0,0 +1,54 @@
...
+    return self.data
+
+def get_kth_last(head, k):
+  if (head == None or k < 0):
+    return None
+  
+  elem = head
+  counter = 0
+  
+  while (elem is not None):
```
jdb: Do you think you can add a __len__ method to the Node?

randomUsernameAndPassword/assignment1/Q2.py
```diff
@@ -0,0 +1,54 @@
...
+
+  def get_next(self):
+    return self.next
+
+  def get_data(self):
+    return self.data
+
+def get_kth_last(head, k):
+  if (head == None or k < 0):
+    return None
```
jdb: returning None is uncommon and is ambiguous as the k-th element could actually be None and you would have no way to distinguish the two.

It is much more common to raise an Exception:

>>> [1,2,3][10]
IndexError: list index out of range

Can you raise an IndexError?

randomUsernameAndPassword/assignment1/Q2.py
```diff
@@ -0,0 +1,54 @@
...
+    else:
+      self.next = None
+
+  def get_next(self):
+    return self.next
+
+  def get_data(self):
+    return self.data
+
+def get_kth_last(head, k):
```
jdb: Do you think you can get the bracket notation to work?

list = Node([4, 2, 5, 6, 9, 8])
list[-2] -> 6

Bracket notation for ordered container is more common, it is easier for your users who do not have to learn a new API for a new container.

To make the bracket notation work, you need to implement __getitem__ on the Node class.

jdb (changes requested):

randomUsernameAndPassword/assignment1/Q2.py
randomUsernameAndPassword: Really nice improvement, thank you!

randomUsernameAndPassword/assignment1/Q2.py
randomUsernameAndPassword: Fixed

randomUsernameAndPassword/assignment1/Q2.py
randomUsernameAndPassword: Fixed

randomUsernameAndPassword/assignment1/Q2.py
randomUsernameAndPassword: Fixed

randomUsernameAndPassword/assignment1/Q1.py
randomUsernameAndPassword: Fixed

randomUsernameAndPassword/assignment1/Q2.py
```diff
@@ -1,41 +1,44 @@
...
+            raise IndexError()
+        if (index < 0):
+            index = index + length
+        
+        elem = self
+        for i in range(index):
+            elem = elem.next
+        return elem
+            
+    def len(self):
```
jdb: You can __len__ to make your class compatible with the builtin len() Python function.

jdb (approved):

PR #6, opened 2018-05-22
SilverSolver: both exercises are done

SilverSolver:
Exercises + some tests

### DIFF
--- a/SilverSolver/assignment1/Q1.py
+++ b/SilverSolver/assignment1/Q1.py
@@ -1,0 +1,49 @@
+def are_words_anagram(s1, s2, case_sensitive=False, return_dicts=False):
+    alphabet = "abcdefghijklmnopqrstuvwxyz"
+    if not case_sensitive:
+        s1 = s1.lower()
+        s2 = s2.lower()
+    else:
+        alphabet += alphabet.upper()
+    alphabet += "0123456789"
+    s1 = s1.strip()         # Ignoring spaces
+    s2 = s2.strip()
+
+    char_counter_1 = dict.fromkeys(list(alphabet), 0)
+    char_counter_2 = char_counter_1.copy()
+    for c in s1:
+        if c in char_counter_1.keys():
+            char_counter_1[c] += 1
+    for c in s2:
+        if c in char_counter_2.keys():
+            char_counter_2[c] += 1
+
+    if return_dicts:
+        return (char_counter_1, char_counter_2)
+    else:
+        return char_counter_1 == char_counter_2
+
+def are_sentences_anagram(s1, s2, case_sensitive=False):
+    s1 = s1.strip()
+    s2 = s2.strip()
+    words1 = s1.split()
+    words2 = s2.split()
+    if len(words1) != len(words2):
+        return False
+    else:
+        words_counter_1 = dict()
+        words_counter_2 = dict()
+        for i in range(len(words1)):
+            w1, w2 = are_words_anagram(words1[i], words2[i], \
+                     case_sensitive=case_sensitive, return_dicts=True)
+            w1 = frozenset(w1.items())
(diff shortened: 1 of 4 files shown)


SilverSolver/assignment1/Q1.py
```diff
@@ -0,0 +1,49 @@
...
+    if not case_sensitive:
+        s1 = s1.lower()
+        s2 = s2.lower()
+    else:
+        alphabet += alphabet.upper()
+    alphabet += "0123456789"
+    s1 = s1.strip()         # Ignoring spaces
+    s2 = s2.strip()
+
+    char_counter_1 = dict.fromkeys(list(alphabet), 0)
```
jdb, line 12: I like this approach with the dict counting the occurences. There is such a a structure already in the standard library: it returns the dict of occurences when you give it an iterable.
collections.Counter

Can you use it?

https://docs.python.org/3/library/collections.htm

SilverSolver/assignment1/Q1.py
```diff
@@ -0,0 +1,49 @@
+def are_words_anagram(s1, s2, case_sensitive=False, return_dicts=False):
+    alphabet = "abcdefghijklmnopqrstuvwxyz"
```
jdb, line 2: defining the alphabet yourself is prone to mistakes, and is likely to diverge from other alphabet definition.
Can you use the character classes "non-whitespace" found in the re module? The symbol for the character class is \W
https://docs.python.org/3/library/re.html

SilverSolver/assignment1/Q1_test.py
```diff
@@ -0,0 +1,29 @@
+import unittest
```
jdb, line 1: Excellent !

SilverSolver/assignment1/Q2.py
```diff
@@ -0,0 +1,38 @@
...
+
+def return_k_th_from_end_element(k, start_node):
+    list_len = 0
+    current_node = start_node
+    while (current_node):
+        list_len += 1
+        current_node = current_node.child
+    element_num = list_len - k
+    if element_num < 0 or element_num > list_len:
+        print("wrong k")
```
jdb, line 14: It is more useful and more common to raise Exception in case of problems.

For example:
>>> [1,2,3][10]
IndexError: list index out of range

Do you think you can raise an Index error instead of the print side effect?

SilverSolver/assignment1/Q2.py
```diff
@@ -0,0 +1,38 @@
+class single_linked_node:
+    def __init__(self, value=None, child=None):
+        self.child = child
+        self.value = value
+
+def return_k_th_from_end_element(k, start_node):
```
jdb, line 6: A much Pythonic interface for ordered container is the brackets notation

start_node[-k] would return the node, or directly the value.

Do you think you can make thsi work? This requires implemented the __getitem__ method on the single_linked_node.

SilverSolver/assignment1/Q2.py
```diff
@@ -0,0 +1,38 @@
...
+            current_node = current_node.child
+
+if __name__ == "__main__":
+    # Demonstration of the functional
+    test_list = [single_linked_node(0, None)]
+    for i in range(1, 10):
+        test_list.append(single_linked_node(i, test_list[-1]))
+    print("we generated list 0 <- 1 <- 2 <- 3 <- 4 <- 5 <- 6 <- 7 <- 8 <- 9 for example")
+
+    current_node = test_list[-1]
```
jdb, line 32: Can you provide a constructor for the linked list? I mean a module function that takes an iterable of values and returns a linked list that holds all the values? I suggest the name from_iterable for your function.

This would make creating a linked list much easier for the users of this data structure.

jdb (changes requested):

PR #8, opened 2018-05-23
solutions

evinpinar:


epinar/assignment1/q1.py
```diff
@@ -0,0 +1,30 @@
+
+def isAnagram(str1, str2, caseSensitive = False):
+	if not caseSensitive:
+		str1 = str1.lower()
+		str2 = str2.lower()
+
+	return (sorted(list(str1)) == sorted(list(str2)))
```
jdb: The normalization logic is duplicated quite a few times: can you put it in a function and reuse that function for both strings in both exercices?

epinar/assignment1/q2.py
```diff
@@ -0,0 +1,63 @@
...
+				return None
+			stop = stop.next
+
+		while stop.next:
+			get = get.next
+			stop = stop.next
+
+		return get.data
+
+def main():
```
jdb: Please create unit tests, I think the way you create the linked list, you revert the order of the input list.

jdb (changes requested):

epinar/assignment1/q2.py
evinpinar: Unit tests created, not sure why you thought like that.

epinar/assignment1/q1.py
evinpinar: Yes in a similar way to your answer.

evinpinar (commented):
by the way, I believe that the linked lists return the last element from 0th index, the last element is 0, the one before is "last 1st"... etc.

epinar/assignment1/q2.py
```diff
@@ -0,0 +1,65 @@
...
+            self.head = new_node
+        else:
+            self.curr.next = new_node
+
+        self.curr = new_node
+        self.size += 1
+
+        return True
+
+    def print(self):
```
jdb, line 27: If you define the __repr__ method, then you can print a LinkedList without having to learn the method.

>>> ll = LinkedList()
>>> ll.add(1), ll.add(2)

Current solution:
>>> print ll.print()

If ll.print is renamed ll.__repr__
>>> print ll

jdb (approved):

PR #9, opened 2018-05-24
Silver solver : corrections

SilverSolver:
I read your recommendations and made corrections

### DIFF
--- a/SilverSolver/assignment1/Q1.py
+++ b/SilverSolver/assignment1/Q1.py
@@ -1,22 +1,15 @@
+from collections import Counter
+import re
+
 def are_words_anagram(s1, s2, case_sensitive=False, return_dicts=False):
-    alphabet = "abcdefghijklmnopqrstuvwxyz"
     if not case_sensitive:
         s1 = s1.lower()
         s2 = s2.lower()
-    else:
-        alphabet += alphabet.upper()
-    alphabet += "0123456789"
-    s1 = s1.strip()         # Ignoring spaces
-    s2 = s2.strip()
+    s1 = re.sub(r'[\W\s]', '', s1)
+    s2 = re.sub(r'[\W\s]', '', s2)
 
-    char_counter_1 = dict.fromkeys(list(alphabet), 0)
-    char_counter_2 = char_counter_1.copy()
-    for c in s1:
-        if c in char_counter_1.keys():
-            char_counter_1[c] += 1
-    for c in s2:
-        if c in char_counter_2.keys():
-            char_counter_2[c] += 1
+    char_counter_1 = Counter(s1)
+    char_counter_2 = Counter(s2)
 
     if return_dicts:
         return (char_counter_1, char_counter_2)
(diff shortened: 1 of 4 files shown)


SilverSolver/assignment1/Q1.py
```diff
@@ -0,0 +1,41 @@
...
+
+    if return_dicts:
+        return (char_counter_1, char_counter_2)
+    else:
+        return char_counter_1 == char_counter_2
+
+def are_sentences_anagram(s1, s2, case_sensitive=False):
+    s1 = re.sub(r'[\W]', ' ', s1)
+    s2 = re.sub(r'[\W]', ' ', s2)
+    words1 = s1.split()
```
jdb: have you seen re.split()? I think it does what you need in one line.

SilverSolver/assignment1/Q2.py
```diff
@@ -0,0 +1,44 @@
...
+    def __getitem__(self, k):
+        k = -k
+        list_len = 0
+        current_node = self
+        while (current_node):
+            list_len += 1
+            current_node = current_node.child
+        element_num = list_len - k
+        if element_num < 0 or element_num > list_len:
+            raise IndexError('wrong k')
```
jdb: Think about the poor developer that needs to make the software work for tomorrow and software breaks all the time, and the only way they can debug is via the logs, and the logs says 'wrong k' :)
Here the problem is that the error message does not define 'wrong'. IMHO, here is a better message for them:

raise IndexError("Index '%s' is bigger than the length of the list '%s'" % (k, list_len))

SilverSolver/assignment1/Q2.py
```diff
@@ -0,0 +1,44 @@
...
+        else:
+            current_num  = 0
+            current_node = self
+            while(current_node):
```
jdb: Mmmh, interesting, Here, you start over, once more, from the beginning of the list.  SO what if the the list is very very long, you see where I am going?
Can we discuss a different implementation of that method that allows to only go through the list only once?

jdb (changes requested):

SilverSolver/assignment1/Q2.py
SilverSolver: I could keep in the memory k previous visited nodes of the singly linked list and return the k-th previous visited node when I reach the end of the list.
But it will consume more memory. So I don't know is it ok?

SilverSolver/assignment1/Q2.py
```diff
@@ -0,0 +1,57 @@
...
+
+class single_linked_node:
+
+    def __init__(self, value=None, child=None):
+        self.child = child
+        self.value = value
+
+    def __getitem__(self, k, one_pass=True):
+        list_len = 0
+        visited_nodes = deque(maxlen=-k + 1)
```
jdb: Pretty cool, deque is a nice container. You are really close to the one pass, and no extra memory: can you have a solution that does not keep all the intermediary nodes in the deque?

:)

jdb (changes requested):

SilverSolver/assignment1/Q2.py
```diff
@@ -1,40 +1,23 @@
...
-    def __getitem__(self, k, one_pass=True):
-        list_len = 0
-        visited_nodes = deque(maxlen=-k + 1)
+    def __getitem__(self, k):
+        current_pos = 0
+        k_th_back_node = self
         current_node = self
         while (current_node):
-            # if we want optimize by time and do one_pass through the
-            # linked list, we should keep k last visited nodes
```
jdb: I agree with you Laida, doing only one pass is not an optimization if the single pass does twice more work. The one pass is kust an alternative, thanks for pointing it

jdb (approved):

PR #10, opened 2018-05-26
JD's linked list.

jdb:
My version of the linked list question. What do you think?

Please point the lines that are not clear, I will try to answer the comments in the github UI and in comments in the code in the same pull request.

### DIFF
--- a/jdb/assignment1/q2-linkedlist.py
+++ b/jdb/assignment1/q2-linkedlist.py
@@ -41,8 +41,14 @@ class TestLinkedList(unittest.TestCase):
     self.assertEqual([n.value for n in LinkedList(list_)], list_)
 
   def test_len(self):
-    for l in ([1], [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]):
-      self.assertEqual(len(LinkedList(l)), len(l))
+    """len() of a LinkedList must be equal to the len of the input list."""
+    # for 2 sample lists, I test that the len of the list is the len
+    # of the LinkedList that is constructed with the list.
+    l1 = [1]
+    self.assertEqual(len(LinkedList(l1)), len(l1))
+    l2 = [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
+    self.assertEqual(len(LinkedList(l2)), len(l2))
+       
 
   def test_forward(self):
     ll = LinkedList(range(10))
--- a/jdb/assignment1/q2-linkedlist.py
+++ b/jdb/assignment1/q2-linkedlist.py
@@ -2,29 +2,52 @@ class Node:
   def __init__(self, value=None):
     self.value, self.next = value, None
 
+# This implementation uses the iterator a lot, it is not better in
+# terms of complexity:
+# pros: shorter methods, more consistent with other Python containers list, dicts...
+# cons: confusing to the reader unfamiliar with iterators
+#
+# resource on Python's iterators:
+# https://www.programiz.com/python-programming/iterator
+# http://www.diveintopython3.net/iterators.html
 
 class LinkedList:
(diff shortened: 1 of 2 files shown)


jdb/assignment1/q2-linkedlist.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class LinkedList(object):
+  def __init__(self, original_list):
+    current = self.head = Node(original_list[0])
+    for value in original_list[1:]:
+      current.next = Node(value)
+      current = current.next
+
+  def __iter__(self):
```
SilverSolver, line 22: Thank you, it was useful for me to look at this iterator implementation. I didn't do it by myself. Also I found useful this material about iterators in python, which also may be interesting for other trainee if they also want to go a bit deeper into the matter:

https://www.programiz.com/python-programming/iterator

At least it was interesting for me..

jdb/assignment1/q2-linkedlist.py
```diff
@@ -0,0 +1,58 @@
...
+
+class TestLinkedList(unittest.TestCase):
+
+  def test_construct_iter(self):
+    list_ = [1,2,3,4,5,6,6]
+    self.assertEqual([n.value for n in LinkedList(list_)], list_)
+
+  def test_len(self):
+    for l in ([1], [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]):
+      self.assertEqual(len(LinkedList(l)), len(l))
```
SilverSolver: To be honest, I didn't understand what is going on here :thinking:

jdb/assignment1/q2-linkedlist.py
jdb: the iterator is tricky: there are two conditions for a class to work correctly with for loops:
1. __iter__ must be defined
2. __iter__ must return an iterator object, that is a object that has a __next__ method and the __next__ method must raise a StopIteration exception to signal the end of the loop

In my example: the __iter__ method is using yield, so automatically it is an iterator.

Thanks for the link Laida, this one is a core page.

jdb/assignment1/q2-linkedlist.py
jdb: Let me try to clarify, I added comments in the code, please take a look.

jdb/assignment1/q2-linkedlist.py
```diff
@@ -0,0 +1,58 @@
...
+      current = current.next
+
+  def __iter__(self):
+    current = self.head
+    while current:
+      yield current
+      current = current.next
+
+  def __len__(self):
+    for index, _ in enumerate(self): pass
```
Sridha2412: I understand what this is doing but why is it better in terms of complexity?

Sridha2412 (changes requested):
I can't understand the complexity of this solution

jdb/assignment1/q2-linkedlist.py
jdb: I assume you ask "better in terms of complexity" compared to the solution you wrote.
Well, it is not better: this solution finds the kth element in 2 steps:
1. find the length: O(n)
2. hop from head len - k times (k is positive): O(n)
The complexity is O(n): this is typical from a linked list, we can not really beat that.

Compared to the solution you wrote, I made len its won method, as it is always useful, and integrated "k from the last" in the bracket notation, because the Python list does the same.

jdb:
This solution has the same complexity as your solution Sridha. Your solution has the benefit to be easier to read to anyone unfamiliar with the Python iterators. I added more explanation and resources as comments in the code. Please take a look.

Sridha2412:
Thanks, I understand it now. :)

Sridha2412 (approved):

PR #11, opened 2018-05-26
JD's anagram solution.

jdb:
Here is a take at this problem. Let me know if there is anything we can discuss.


Sridha2412:
Thanks for the solution. It was helpful for me to look at a Counter implementation :)

Sridha2412 (approved):
Thanks for the solution. It was helpful for me to look at a Counter implementation :)

PR #13, opened 2018-06-03
Sridha2412 2

Sridha2412:


sridha2412/assignment2/q2.py
```diff
@@ -0,0 +1,60 @@
...
+root.left.right = Node(5)
+root.right.right = Node(8)
+root.left.left.left = Node(1)
+root.left.left.right = Node(6)
+
+node1 = 6
+node2 = 5
+
+# check both nodes are present in tree
+if not find_node(root, node1) or not find_node(root, node2):
```
jdb: Can you implement the check in the same run/pass as in the find_lca method?
I was confused by fund_node, with such a name, we would expect the method to return the node, but it is a predicate (a function that returns a bool). It is common to name predicates is_<adjective>. I suggest "is_present". 

A more common way to write find_node following the Python convention is to implement __getitem__ like on the LinkedList for the previous exercice, and raise a KeyError when the value is not found.

sridha2412/assignment2/q1.py
```diff
@@ -0,0 +1,36 @@
...
+
+# node class`
+class Node:
+    # constuctor to initlalize a node object
+    def __init__(self, data):
+        self.data = data
+        self.left = None
+        self.right = None
+
+# if key is present, print ancestors else return false
```
jdb: Please set these comments as "docstrings":
https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

def print_ancestors(root, key):
  """If key is present, print ancestors else return false."""

sridha2412/assignment2/q1.py
```diff
@@ -0,0 +1,36 @@
...
+def print_ancestors(root, key):
```
jdb: Please document the functions signatures with Python's type annotations.
1. this makes it easier for the reader,
2. this makes you think about what makes sense for good types for your module.

I suggest you split the print_ancestors function in two:
1. one function that returns the list of ancestors,
2. one function that simply prints it.

The advantage of having a function that returns a list is that it is much more useful, instead of being scoped to simply printing. 

Even in our limited use case, it is much easier to do unit test on a list return value than on a values being printed. Testing that function prints what we expect is possible but less straightforward.

sridha2412/assignment2/q1.py
```diff
@@ -0,0 +1,36 @@
+# print ancestors of given key in a binary tree
```
jdb, line 1: I suggest you do unit test :) At the last mentors meeting, it was decided that unit tests will only be required for the next assignment, but I suggest you still do it for this assignment.

jdb (changes requested):

sridha2412/assignment2/q1.py
```diff
@@ -2,29 +2,41 @@
...
+    """ if key is present, return list of ancestors else return false """
+    # base cases
     if root is None:
         return False
-
     if root.data == key:
         return True
 
+    # store all nodes in list
+    ancestors.append(root.data)
```
jdb: Just note that the exercice request that you do not store additional nodes and values, and list.append is storing the values here :)

sridha2412/assignment2/q1.py
```diff
@@ -2,29 +2,41 @@
...
 
 
+#       7
+#      / \
+#     3   4
+#    / \   \
+#   2   5   8
+#  / \
+# 1   6
+
```
jdb: Please remove the code that executes on import, you can move it into a unit test.

sridha2412/assignment2/q1.py
```diff
@@ -33,4 +45,37 @@ def print_ancestors(root, key):
...
+class Test(unittest.TestCase):
+    def setUp(self):
+        #       7
+        #      / \
+        #     3   4
+        #    / \   \
+        #   2   5   8
+        #  / \
+        # 1   6
+        self.root = Node(7)
```
jdb, line 42: An alternative to the procedural/imperative style of creating a tree is the constructors of the tree that is driven by the input. In the solution I suggest, leftness or the rightness of the subtrees is decided by the presence of a element in the second or third position of the input tuple. 

Both solution works, the solution that is driven by the data is more flexible as you can generate trees programmatically by simply creating a function that returns the right tuple. In the case of procedural code like below, you would need to generate, not a tuple but generate the code itself. Code generation is fun but it is another complex software problem.

sridha2412/assignment2/q2.py
```diff
@@ -2,44 +2,51 @@
...
-    if root.data == node or find_node(root.right, node) or find_node(root.left, node):
+    if root.data == key or is_present(root.right, key) or is_present(root.left, key):
         return True
 
     return False
 
-# method to find least common ancestor of two given nodes
-# note: if one key is ancestor of another, it returns the key
-def find_lca(root, node1, node2):
+def find_lca(root: Node, node1: int, node2: int) -> int:
```
jdb: I do not understand this method, sorry, can you expand the docstring or the comments to explain the approach?

jdb (changes requested):
Thanks for the unit tests and the types ! Please explain the lca method. I will take another look tonight but at this point, I do not really understand it.

sridha2412/assignment2/q1.py
Sridha2412: I thought that constraint was for the second question. 😅  I can't figure out a way to return a list without storing the values, the only way  I can think of is using iterators but that would completely refactor my code. Do you have any other suggestions or do I fully refactor the code?

sridha2412/assignment2/q1.py
Sridha2412: I agree, my approach isn't flexible at all. I only did so to quickly test it 😅

sridha2412/assignment2/q2.py
Sridha2412: I have added more comments to this method, hope that explains it better

sridha2412/assignment2/q1.py
```diff
@@ -8,24 +8,26 @@ def __init__(self, data: int):
         self.left = None
         self.right = None
 
-def find_ancestors(root: Node, key: int) -> bool:
+def find_ancestors(root: Node, key: int, ancestors: list) -> bool:
```
jdb: It is confusing, the return value is a boolean and the docstrings states the function returns the list of ancestors. 

It would be easier for the user of your API that they do not have to create the list themselves and the function creates it. 

This is a simple change:

def find_ancestors(root: Node, key: int, ancestors: list = None) -> bool:
  if ancestors is None:
    ancestors = []

sridha2412/assignment2/q1.py
```diff
@@ -8,24 +8,26 @@ def __init__(self, data: int):
...
     """ if key is present, return list of ancestors else return false """
     # base cases
     if root is None:
         return False
     if root.data == key:
         return True
 
+    # store all nodes in list
```
jdb: This comment is not useful: here you rewrite the code in plain english. What is usually more interesting as a comment is not the "what" but the "why".

If the why is obvious, no need for a comment.

sridha2412/assignment2/q1.py
```diff
@@ -8,24 +8,26 @@ def __init__(self, data: int):
         self.left = None
         self.right = None
 
-def find_ancestors(root: Node, key: int) -> bool:
+def find_ancestors(root: Node, key: int, ancestors: list) -> bool:
     """ if key is present, return list of ancestors else return false """
     # base cases
     if root is None:
         return False
```
jdb: In Python, key is not present is spelled KeyError. Instead of returning values that means "something went wrong", the convention is to raise an exception. So instead of:

return False

you write:

raise KeyError('value %s not found in the tree' % key)

sridha2412/assignment2/q1.py
jdb: Have you thought about attaching the method to the class itself. It really looks ready for it: replace 'root' with 'self' and add two more spaces in front of each lines.

jdb (changes requested):

jdb (approved):

PR #14, opened 2018-06-03
Epinar 2

evinpinar:


epinar/assignment2/ancestors.py
```diff
@@ -0,0 +1,113 @@
...
+		self.right = None
+
+# Initiates a tree with an input value. Keeps the nodes in a dictionary for fast reach.
+# Each data(key) value should be unique.
+class BinaryTree:
+	def __init__(self, data):
+		self.tree = {}
+		node = TreeNode(data)
+		self.head = node
+		self.tree[data] = node
```
jdb, line 14: I feel like this is cheating :) It should be a tree, not a dict. The constraints of keys being unique is too strong a requirement, then this severely limits the usefulness of the tree.

1. Please list at least two additional difficultries with this approach,
2. Please do without the self.tree index, to make sure we solve the problem of a tree with its inherent complexity and structure.

epinar/assignment2/ancestors.py
```diff
@@ -0,0 +1,113 @@
+class TreeNode:
```
jdb, line 1: I think you can simplify the design of this module by merging the TreeNode and the BinaryTree.
The tree is an inherently recursive data structure: can you try to attach the interesting methods to TreeNode and each start operating from the local node?

epinar/assignment2/ancestors.py
```diff
@@ -0,0 +1,113 @@
...
+	def __init__(self, data):
+		self.data = data
+		self.left = None
+		self.right = None
+
+# Initiates a tree with an input value. Keeps the nodes in a dictionary for fast reach.
```
jdb, line 7: Please attach the comments as Python docstrings.
https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

epinar/assignment2/ancestors.py
jdb, line 1: Can you try to merge the TreeNode and BinaryTree classes?
The split between the 2 classes is not very clear (except for holding the self.tree cache, please see my comment below). The two classes are interdependent, and the contract for each class is not clearly defined which is a source of later problem and confusion. 

Another name for this pattern is "inappropriate intimacy":
https://blog.codinghorror.com/code-smells/

epinar/assignment2/ancestors.py
```diff
@@ -0,0 +1,113 @@
...
+	def print_dict(self):
+		for k, v in self.tree.items():
+			print("node: ", v.data)
+			if v.right:
+				print(" r: ", v.right.data)
+			if v.left:
+				print(" l: ", v.left.data)
+
+	# Prints the tree in order.
+	def print(self):
```
jdb, line 38: Please use the conventional name for this method: __repr__.
This simplifies the tasks for the user: to get a reasonable representation, they can simply get it with the builtin print, instead need to use: 
>>> print my_node.print()

epinar/assignment2/ancestors.py
```diff
@@ -0,0 +1,113 @@
...
+	def _print(self, curr):
+		if not curr:
+			return
+		print(curr.data)
+		self._print(curr.left)
+		self._print(curr.right)
+
+	# Returns the ancestors ordered from lowest to highest in tree.
+	def ancestors(self, data):
+		self.a = []
```
jdb, line 50: Mmh, I do not understand the role of this a. The name is confusing, please avoid one-letter names except for iterators and counters:
https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming

Here I think you attaching to the tree data whose lifetime is the run of the ancestors() method. I do not think you should be attaching the state of the ancestors computation to the tree itself. For instance, this make the ancestors() method incompatible with concurrent programming where, for example, 2 threads would be stepping on each other toes by both appending to the same self.a variable.

I think you are solving the problem of sharing data between the runs of a recursive function. In this case, a is called in general, an accumulator. Try to be explicit instead of implicit: here, pass the accumulator via the function arguments, instead of passing some arguments with the function calls, and others via the root object (this solution has the benefit to solve the problem of the concurrency).

epinar/assignment2/ancestors.py
```diff
@@ -0,0 +1,113 @@
...
+		p.left = node
+		self.tree[data] = node
+	def add_right(self, parent, data):
+		p = self.tree[parent]
+		node = TreeNode(data)
+		p.right = node
+		self.tree[data] = node
+
+	# Prints in dictionary order.
+	def print_dict(self):
```
jdb, line 29: Please use the conventional name for this method: __repr__.
This simplifies the tasks for the user: to get a reasonable representation, they can simply get it with the builtin print, instead need to use: 
>>> print my_node.print()
>>> print my_node

When we use several dozens Python modules in a project, if we do not have to learn each of their peculiar methods names, we are more productive.

jdb (changes requested):

jdb (approved):

epinar/assignment2/ancestors.py
evinpinar: Will do for the next assignments!

epinar/assignment2/ancestors.py
evinpinar: Totally makes sense, I will consider it next time!

epinar/assignment2/ancestors.py
evinpinar: I will use that naming from now on!

epinar/assignment2/ancestors.py
evinpinar: Here, instead of printing the nodes, I created this variable to accumulate them. Then, I learnt a nice way to do that in python: by yielding. I will try to apply that wherever I see needed next times.

epinar/assignment2/ancestors.py
evinpinar: 1. I thought the assignment question itself required unique keys. As you mentioned, it restricts the usefulness. As well, first, it will be hard to manipulate the tree, for example removing the head, or adding an additional node in between some nodes. Second, it makes it impossible(very messy and weird) to define the structure of a node, being a tree at the same time, such as in your solution. Is that right?
2. I think I understand the overall point here, I will be applying the idea from now on wherever needed.

evinpinar (commented):
Thanks for the review!

PR #15, opened 2018-06-04
assignment 2

crossopt:


crossopt/assignment2/treenode.py
```diff
@@ -0,0 +1,82 @@
...
+        ans = []
+        cnt = self.parent
+        while cnt:
+            ans.append(cnt.val)
+            cnt = cnt.parent
+        return ans
+
+    # given two nodes in subtree, returns lca-node
+    def lca(self, v1, v2):
+        if v1 not in self or v2 not in self:
```
jdb: this check is costly: it is O(nodes), the tree is traversed twice here. This is the slowest part of your algorithm as height() below is in O(height) ~ O(ln(nodes))

I suggest you do not put this check at the start of  this method and provide this check as a separate method, the user gets the choice to use it and pay the penalty.

crossopt/assignment2/treenode.py
```diff
@@ -0,0 +1,82 @@
...
+    def height(self):
+        ans = 0
+        cnt = self.parent
+        while cnt:
+            ans += 1
+            cnt = cnt.parent
+        return ans
+
+    # returns list of ancestors of current node
+    # TODO - this and height look similar
```
jdb: I agree with you comment: height is equal to the length of the ancestors list. There is a big difference though, your height() function does not store additional nodes, which is part of the requirement in the exercice.

crossopt/assignment2/treenode.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class TreeNode:
+    def __init__(self, val, par=None):
+        self.left = None
+        self.right = None
+        self.val = val
+        self.parent = par
+
+    def __iter__(self):
```
jdb, line 11: Cool, I did the same iterator!

crossopt/assignment2/treenode.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+
+
+class TreeNode:
+    def __init__(self, val, par=None):
+        self.left = None
+        self.right = None
+        self.val = val
+        self.parent = par
```
jdb, line 9: why storing the parent and not the height?
Do dyou see a difference between the two?

jdb (changes requested):

crossopt/assignment2/treenode.py
crossopt: Storing the parent rather than the height makes iterating upwards possible, which in turn makes getting all of the ancestors of a node more simple
Also, storing parents instead of heights hypothetically makes linking trees easier

crossopt/assignment2/treenode.py
crossopt: OK.
But why is O(height) ~ O(ln(nodes))?
We don't know that our tree is balanced, so we could theoretically have just one long rope of length nodes and it would be considered a correct binary tree

jdb (approved):

PR #16, opened 2018-06-05
Silver solver 2

SilverSolver:
I made baseline solutions for the second assignment and tests for them.
I'll try to think how to improve 2nd solution (get lowest common ancestor) with no using of additional space for lists of common ancestors, but for now I still didn't find out how to do this


SilverSolver/assignment2/binary_tree.py
```diff
@@ -0,0 +1,61 @@
...
+        self_anc = self.get_ancestors(value=False)
+        another_anc = another_node.get_ancestors(value=False)
+        low_c_anc = None
+        while(self_anc and another_anc):
+            if self_anc[-1] == another_anc[-1]:
+                low_c_anc = self_anc[-1]
+                self_anc = self_anc[:-1]
+                another_anc = another_anc[:-1]
+            else:
+                if value and low_c_anc:
```
jdb: I suggest you go with the value=True everywhere.

SilverSolver/assignment2/binary_tree.py
```diff
@@ -0,0 +1,61 @@
...
+            return self.parent.get_ancestors(prev_ancestors=prev_ancestors, value=value)
+
+    def get_lowest_common_ancestor(self, another_node, value=True):
+        self_anc = self.get_ancestors(value=False)
+        another_anc = another_node.get_ancestors(value=False)
+        low_c_anc = None
+        while(self_anc and another_anc):
+            if self_anc[-1] == another_anc[-1]:
+                low_c_anc = self_anc[-1]
+                self_anc = self_anc[:-1]
```
jdb: The bracket notation with the colon creates a copy of the list, the copy can get costly, an integer index would avoid the copy.
There is the zip() builtin iterates on two list at the same time, can you use that?

SilverSolver/assignment2/binary_tree.py
```diff
@@ -0,0 +1,61 @@
...
+            return low_c_anc
+
+class BinaryTree:
+
+    def __init__(self, iterable=None):
+        if iterable and len(iterable) > 0:
+            iterable = list(map(BinaryNode, iterable))
+            self.root = iterable[0]
+            current_nodes = [self.root]
+            iterable  = iterable[1:]
```
jdb: to avoid the copy, you might want to use a deque, or revert the list to be able to use pop.
list.pop() is O(1) while list[1:] is O(n)

SilverSolver/assignment2/binary_tree.py
```diff
@@ -0,0 +1,61 @@
...
+                if value and low_c_anc:
+                    return low_c_anc.value
+                else:
+                    return low_c_anc
+        if value and low_c_anc:
+            return low_c_anc.value
+        else:
+            return low_c_anc
+
+class BinaryTree:
```
jdb, line 40: What are  the advantages of using two classes? I found it simpler to use only a class Node with a recursive constructor. In any case, there is no important differences between the two approach.

SilverSolver/assignment2/binary_tree.py
```diff
@@ -0,0 +1,61 @@
+class BinaryNode:
+
+    def __init__(self, value, parent=None, child1=None, child2=None):
+        self.value  = value
+        self.parent = parent
+        self.child1 = child1
+        self.child2 = child2
+
+    def get_ancestors(self, prev_ancestors=None, value=True):
```
jdb, line 11: I see you are supporting both returning the values and returning the nodes. The exercise is not saying explicitly which way we should go, so choose one direction, and document it. This will simplify the code a bit.
Myself, I went with returning the *values* in my solution.

jdb (changes requested):

SilverSolver/assignment2/binary_tree.py
SilverSolver: I suggest that two **different** nodes can store the same values, but always have different addresses in RAM/pointers etc.
I motivate this approach by the fact that nodes with the same value can have different parents and children (at least I see no obstacles to this, because value, children and parents are different and independent fields of the node object in my implementation), so they can have different positions in the tree; but nodes with the same address in RAM are factually the same object with the same position and all and should be considered as the same node.
If I understood the assignment correctly, that in the **first** part I should return the **values** of ancestors of the node. But in the **second** part I should return the value of **common ancestor** . So I thought that I should find the _object_, which is the common ancestor of two nodes in the tree structure. That's why I thought that I should have some option to store and comparing ancestor _objects_ itself (i.e. pointers to objects) - not only their _values_ - to see the difference between possibly different ancestors with the same value. And that's why I added the possibility to get pointers to ancestors (if value=False), not only their values (which is going on if value=True).

Let's consider a little example.
0
| \
1 1
|  |
2 2
If I will use all my code with value=True at this example, it will say that the lowest common ancestor for least nodes (with values 2 and 2) is "1", which have no sense in my opinion, as I tried to explain before. But if I will use all my code as I use it now (i.e. with value=False when I need this), it will say that the lowest common ancestor is "0", which seems reasonable.

Sorry if this explanation seems too excess to you. My English isn't very good, that's why I tried express an idea in as much detail as possible, to neutralize possible misunderstanding. I hope that it is clear now, why I should have option to use value=False!

SilverSolver/assignment2/binary_tree.py
SilverSolver: I'm not sure what do you mean by "document it". Also look to my answer to your first comment about why I need to supporting the both returning the values and the objects.

SilverSolver/assignment2/binary_tree.py
SilverSolver: I supposed that class BinaryTree will be used to construct tree itself, and BinaryNode will contain information about the particular node. So they have different meaning, in my opinion and that's why can be divided. If you are sure that it is bad, I can re-design this part of the program in one of the following ways:
a) make one function "from_iterable" with the same meaning as constructor of the class BinaryTree
b) turn the constructior of the class BinaryTree into some method of BinaryNode
But I'm not sure whether it is worth the time or not.

SilverSolver/assignment2/binary_tree.py
SilverSolver: Oh, I see now at your variant of solution that you made assumption that values held by the nodes are unique! That's where point of our divergence is. :smile:

SilverSolver/assignment2/binary_tree.py
jdb: A designer knows he has achieved perfection not when there is nothing left to add, but when there is nothing left to take away.
  -- Antoine de Saint-Exupery 

Both solutions works, no need to rewrite, it would be a waste of time, I just want to share a few things, I feel the BInaryTree constructor is not needed as the same feature can be implemented in the Node constructor, depending on whether the optional left and right are provided

**Recursive data structure**
Here is conceptual approach as to why I prefer one class. With BinaryNode and BinaryTree, we get the zoom level locked respectively to the minimum granularity and the maximum granularity. With one class that represent a subtree, we have one view that generalizes both approach. The one class approach is akin to fractals where each node can be zoomed in and present the same structure, and same methods are available.

What if the BinaryTree gets so big that it does not fit in memory anymore? Then you would split it in several subtrees and host each subtree on a different memory server. Would you make each subtree a BinaryNode and keep them in another container? Would you create a third class? I personally find the one class approach more flexible.

**Inappropriate intimacy**
Two objects needs to be compatible, as they use  methods or attributes from each other, this compatiblity needs to be understood and respected for team members wanting to expand this code.
This pattern of Inappropriate intimacy is descrbied here with other software programming wisdom:
https://blog.codinghorror.com/code-smells/
(feel free to read and disagree)

**code is a liability, not a an asset**
Less objects usually means shorter code, low cognitive overhead, less bugs. Less code is also less effort for the team or the reviewers. Code always grow so complex that the velocity in adding new features decrease as the effort to understand the code and the fear of breaking it because we do not understand 100% of it. Here are the kind of features that would need to be implemented where I would add more objects: logging, profiling/statistics, access and authorization checks, parallelism, storage, memory redundancy, performance optimization.
Just to share some of my experience with OOP: I have worked on two projects where the inheritance chain would go more than 10 levels high, it is slow to understand what to modify when adding a feature.

SilverSolver/assignment2/binary_tree.py
jdb: Sorry, I do not see how the answer to the first comment explains the value of supporting two types of return values. Polymorphic methods **on the return value** are quite uncommon. In the context of an inteview, where the whiteboards are not always big, and the time is really limited. I would focus on the minimum amount of code to solve the problem, not adding extra code for extra features.

jdb (approved):

SilverSolver/assignment2/binary_tree.py
SilverSolver: Thanks for detailed explanation. I didn't think about this data structures in terms of scalability. If we consider BinaryNode as a subtree, than BinaryTree is really excessive.

PR #17, opened 2018-06-08
JD's solutions for assignment 2: trees

jdb:
I created 2 versions:
1. Each nodes keeps a pointer to the parent and also keeps as an attribute the depth of each nodes. I suggest you start reading this one, it is simpler and more commented.
2. Nodes really only have left, right and value attribute, so it is a bit harder to implement the exercise efficiently. Take a look if interested.

### DIFF
--- a/jdb/assignment2/tree.py
+++ b/jdb/assignment2/tree.py
@@ -1,0 +1,196 @@
+
+"""A solution for the CodeU Emea assignment #2: binary trees.
+
+There are two important elements, extracted from the exercice, that drove this
+implementation:
+
+1. no mentions of mutations on the tree once it is built, the tree is read-only,
+   this makes it easy and worthwhile to precompute as much things as possible so
+   they are available when needed, and they are only computed once:
+   - Nodes keep a pointer to the parent
+   - Nodes keep a int representing the depth of a Node in the tree.
+
+   If the tree was mutable, we could still make sure mutations would keep these
+   2 signals correct, but it would be more work.
+
+2. the exercise requests no storing of the nodes: this code does not create any
+   list to which nodes get appended, instead, this code uses generators which
+   are comparable to index pointer being moved. They are baked into the Python
+   language with the yield keyword so they are less verbose than in other
+   languages like C++ or Java.
+
+By construction, the two tree properties are guaranteed: connected graph and no
+loops. In addition, one large assumption in this code, is that values held by
+the nodes are unique. The unit test uses ints, there are no 2 nodes with the
+value 2, for example.
+
+Additional note about Python type annotations: Python 3.5 introduced optional
+typing, and types complement unit tests:
+1. they help the readers of the code,
+2. they help the author to design a good consistent API,
+3. type checkers find errors statically, thus saving some class of errors
+   before the code is sent to production.
+4. static analysis helps with refactoring, and refactoring is a fact of life for
+    medium to big sized project.
+We suggest and warmly support the use of Python type annotations :)
+More info at: https://docs.python.org/3/library/typing.html
+
+To run the mypy static type checker (http://mypy-lang.org/):
+$ mypy tree.py
+"""
+
+import unittest
+from typing import Generator, Tuple, TypeVar
+
+T = TypeVar('T')  # the generic type of values held in each node.
+
+class Node:
+  def __init__(self, value : T,
+               left: Tuple = None,
+               right: Tuple = None,
+               parent: 'Node' = None,
+               depth: int = 0):
+
+    self.value = value
+    self.parent = parent
+    self.depth = depth
+    self.left = Node(*left, parent=self, depth=depth+1) if left else None
+    self.right = Node(*right, parent=self, depth=depth+1) if right else None
+
+  def _ancestors(self) -> Generator['Node', None, None]:
+    """Yields the current node, then the ancestor nodes up to the root.
+
+    The first element yield is the node itself (in this sense, the node is
+    considered an ancestor of itself).
+    """
+    yield self
+    if self.parent:
+      yield from self.parent._ancestors()
+
+  def print_ancestors(self):
+    print(' '.join(str(i) for i in self._ancestors()))
+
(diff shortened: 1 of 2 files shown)


jdb/assignment2/tree.py
```diff
@@ -0,0 +1,196 @@
...
+    considered an ancestor of itself).
+    """
+    yield self
+    if self.parent:
+      yield from self.parent._ancestors()
+
+  def print_ancestors(self):
+    print(' '.join(str(i) for i in self._ancestors()))
+
+  def lowest_common_ancestor(self, node1: 'Node', node2: 'Node') -> 'Node':
```
SilverSolver, line 73: I love the way in which this function was written. It is very smart and reasonable. :+1:

jdb/assignment2/tree.py
```diff
@@ -0,0 +1,196 @@
...
+    """Returns the node that is the lowest common ancestor of node1, node2."""
+    deeper, shallower  = ((node2, node1)
+                          if node2.depth > node1.depth else (node1, node2))
+    for _ in range(deeper.depth - shallower.depth):
+      deeper = deeper.parent
```
SilverSolver, line 78: This is a smart way to speed up the function when nodes are at very different depths, I will remember this approach!

jdb/assignment2/tree.py
```diff
@@ -0,0 +1,196 @@
...
+        pass
+  if root.value == value:
+    return root
+  raise KeyError("Value '%s' not found in the tree." % value)
+
+
+def check_loops(value: T,
+                left: tuple = None,
+                right: tuple = None,
+                seen: set =None):
```
SilverSolver, line 107: In my opinion, the interface of this function could be more convenient if it could take a Node as an input.

For example, I don't know how to test the following one-node graph with this function:

```
tricky_node = Node("value")
tricky_node.left = tricky_node
tricky_node.right = tricky_node
tricky_node.parent = tricky_node
# how to use check_loops?..
```

jdb/assignment2/tree_no_parent.py
```diff
@@ -0,0 +1,97 @@
+import unittest
+from typing import Generator, List, Tuple, TypeVar
+
+T = TypeVar('T')
+TAncestors = Tuple[T]
+_StackOfT = List[T]
+
+class Node:
+    def __init__(self, value: T, left = None, right = None):
+        self.left = Node(*left) if left else None
```
SilverSolver, line 10: I'm not sure about what "*" is meaning there. Why not to use just Node(left) etc?

jdb/assignment2/tree_no_parent.py
```diff
@@ -0,0 +1,97 @@
...
+        self.value = value
+        self.right = Node(*right) if right else None
+        
+    def _ancestry_lines(self, stack: _StackOfT = None) -> Generator[_StackOfT, None, None]:
```
SilverSolver, line 14: The name of this function seems a little bit confusing. As I could understand from code, it starts from the root node and returns the lines of descendants, which are using in the function find_ancestors then. Am I understand correctly?

jdb/assignment2/tree.py
jdb: The idea was not to check a node, but to check the input for the constructor of Node (the method Node.__init__). So it does not take a Node as an input but the recursive tuple of tuples data structure that you can find below in the unit test. The idea is that the user checks the input before creating a Node, we expect Node to only be created via the constructor and never mutated after. If this is true Nodes, can never be incorrect.

This is good feedback, I agree the current code as you show in your example allow to create bad nodes. Not only there is a loop, in the example you show, but also, the height is incorrect!

If you want check loops to work on Nodes, here is such a method:

class Node:
  def check_loops(self, seen=None):
    "Returns None if all went well, or raise a ValueError if a loop is detected.."
    if seen is None:
        seen = set()
    if id(self) in seen:
      raise ValueError("Loop detected")
    else:
      seen.add(id(self))
    if self.left:
        self.left.check_loop()
    if self.right:
        self.right.check_loop()

I had a first version that was integrating the loop check in the Node constructor, but the code started to be a bit complicated, messy. I found this version more readable: after all, the "interview question"  does not request checking for loop: so better solve the exact problem first cleanly, then improve on the solution if there is still time.

jdb/assignment2/tree_no_parent.py
jdb: Node(left)  # left will become Node.__init__ value argument.
Node(left)  # left is 'unpacked', its first element becomes __init__'s value, second element becomes left, and third element becomes right. If the initial left is missing a second or third element, then the default value for that argument is used.

The star, applied to a list of a tuple "unpacks" the element of the list or tuple each in the corresponding arguments. See this example:

>>> args = '1st', '2nd', '3rd'
>>> def f(a,b,c):
...   print('All good')
... 
>>> f(*args)
All good
>>> f(args)
TypeError: f() takes exactly 3 arguments (1 given)

The error message is pretty clear: args was passed as the first argument it became 'a', and b and c are missing.

The advantage of this notation is to separate the time where you can create, manipulate, transfer the arguments of a function, and in a second, later time, you can use arguments and apply the function on them. For example, in the unit tests, the recursive tuple of tuple stores the arguments later passed to the Node constructor. It is a powerful feature.

Is this clear?

jdb/assignment2/tree.py
jdb: thanks !

jdb/assignment2/tree_no_parent.py
jdb: This functions yields all possible ancestry lines in the tree. I think you got it right.

Since there is no parent pointer, I need to construct the ancestry lines by remembering all the ancestor of the current node I visit in a stack.

jdb/assignment2/tree_no_parent.py
SilverSolver: Yes, thanks

SilverSolver (approved):

PR #18, opened 2018-06-16
Crossopt 3

crossopt:


crossopt/assignment3/dictionary.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+
+class DictionaryNode:
```
jdb: To clarify the API of this module, I think you can document DictionaryNode private (by prefixing). This helps the reader categorize who uses which objects. I understand in this exercice solution that no-one uses DictionaryNode outside of this module, am I right?

crossopt/assignment3/unittests.py
```diff
@@ -0,0 +1,79 @@
...
+import unittest
+from assignment3.dictionary import Dictionary
+from assignment3.wordsearch import Grid
+from assignment3.wordsearch import findMatch, findMatches
+
+
+class TestDictionary(unittest.TestCase):
+    def test_isprefix(self):
+        dictionary = Dictionary(['CAR', 'CART', 'CARD', 'CAT'])
+        self.assertEqual(dictionary.isPrefix('A'), 0)
```
jdb: I am not sure I understand well: isPrefix returns a boolean and the unit test compares it to 0 or 1. 
Can you confirm the test pass?

crossopt/assignment3/unittests.py
```diff
@@ -0,0 +1,79 @@
...
+        self.assertEqual(dictionary.isWord('car'), 0)
+        self.assertEqual(dictionary.isWord('CARDS'), 0)
+        self.assertEqual(dictionary.isWord(''), 0)
+        self.assertEqual(dictionary.isWord('CARD'), 1)
+        self.assertEqual(dictionary.isWord('CAR'), 1)
+
+
+class TestGrid(unittest.TestCase):
+    def test_getitem(self):
+        array = [['A', 'A', 'R'], ['T', 'C', 'D']]
```
jdb: WIthout changing the implementation of the current Grid, I think you can replace:
 [['A', 'A', 'R'], ['T', 'C', 'D']]

with 

['AAR', 'TCD']

List a list, a string is iterable.

crossopt/assignment3/unittests.py
```diff
@@ -0,0 +1,79 @@
...
+        grid = Grid(array)
+        self.assertEqual(grid.n, 2)
+        self.assertEqual(grid.m, 3)
+        for i in range(grid.n):
+            for j in range(grid.m):
+                self.assertEqual(grid[i][j], array[i][j])
+
+    def test_neighbors(self):
+        grid = Grid([['A', 'A', 'R'], ['T', 'C', 'D']])
+        self.assertSetEqual(set(grid.neighbors(0, 0)),
```
jdb: Why do you transform the list of neighbors into a set?
It looks to me that the neighbors() function returns a stable order, and the set is not needed. Is this correct?

jdb (changes requested):

crossopt/assignment3/unittests.py
crossopt: I've fixed it, however the tests pass on my computer.

crossopt/assignment3/unittests.py
crossopt: Yes, in hindsight that was a bad idea.
When I was writing the tests I assumed that since the only thing I needed from the neighbors function was to return all the cell's neighbors in any order, I didn't need to check in which order they were actually returned

crossopt/assignment3/dictionary.py
```diff
@@ -0,0 +1,43 @@
...
+            self.addWord(word)
+
+    def addWord(self, word):
+        cnt = self.root
+        for letter in word:
+            cnt.add(letter)
+            cnt = cnt.children[letter]
+        cnt.is_terminal = True
+
+    def isPrefix(self, prefix):
```
jdb, line 29: Consider factoring isPrefix and isWord maybe?

jdb (approved):

PR #19, opened 2018-06-17
q1 done

Sridha2412:


sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,124 @@
...
+        current = self.root
+        for char in word:
+            if char not in current.children:
+                return False
+            current = current.children[char]
+        return current.is_word
+
+
+class Grid:
+    def __init__(self, grid: list):
```
jdb: Is it a List[str]? This would be even clearer type annotation wdyt?

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,124 @@
...
+        """ Method to find all possible neighbouring cells given position (x, y) in a grid """
+        # 8 possible positions - vertical, horizontal and diagonal
+        positions = [(x - 1, y - 1), (x - 1, y), (x - 1, y + 1), (x, y + 1), (x + 1, y + 1), (x + 1, y), (x + 1, y - 1), (x, y - 1)]
+        return [pos for pos in positions if 0 <= pos[0] < len(self.grid) and 0 <= pos[1] < len(self.grid[0])]
+
+    def get_word(self, positions: list) -> str:
+        """ Method to find corresponding word given a list of positions in grid """
+        return ''.join(self.grid[pos[0]][pos[1]] for pos in positions)
+
+    def find_words(self, dictionary: Dictionary, position: list) -> list:
```
jdb: There is a handy convention for  reader:  prefix the private methods with an underscore to document. This would help me figure out quickly the internal helper methods from the feature methods.

jdb: If you want to represent two integers, I suggest a tuple instead of a list. It won't change much in this exercices, but conventional, many colleagues expect a list to be of variable len, and somehow indistinguishable elements and expects a tuple to have a fixed number of elements and each element represent something different depending on the position.

Then, in your type, you can write Tuple[int, int].
But you can't write List[int, int], the expectations is that the elements are of the same type in a List, and List expects only one argument.

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,124 @@
...
+        """ Method to return all words from given position """
+        words = []
+
+        word = self.get_word(position)
+        if (not dictionary.is_prefix(word)):
+            return []
+        if (dictionary.is_word(word)):
+            words.append(word)
+
+        neighbours = self.get_neighbours(position[-1][0], position[-1][1])
```
jdb: I do not understand how position[-1][0] can be correct. I thought position was a list of 2 integers?

I now understand reading the usage below: why a list of one tuple where you could simply use the single 2-tuple?

jdb (changes requested):
Well done Sridha !!

sridha2412/assignment3/wordsearch.py
Sridha2412: it is a list of tuples  so it looks something like [(0, 0), (0, 1), ...]
at each iteration, it appends a position to the tuple to avoid using the same cell
It is position[-1][0] because I am accessing the last tuple of the list.

sridha2412/assignment3/wordsearch.py
Sridha2412: it is a list of tuples, I wasn't sure having a tuple of tuples was a good idea

sridha2412/assignment3/wordsearch.py
Sridha2412: This one is a List[int][int] but i have added List[str] where applicable

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,126 @@
...
+        """ Method to find all possible neighbouring cells given position (x, y) in a grid """
+        # 8 possible positions - vertical, horizontal and diagonal
+        positions = [(x - 1, y - 1), (x - 1, y), (x - 1, y + 1), (x, y + 1), (x + 1, y + 1), (x + 1, y), (x + 1, y - 1), (x, y - 1)]
+        return [pos for pos in positions if 0 <= pos[0] < len(self.grid) and 0 <= pos[1] < len(self.grid[0])]
+
+    def _get_word(self, positions: List) -> str:
+        """ Method to find corresponding word given a list of positions in grid """
+        return ''.join(self.grid[pos[0]][pos[1]] for pos in positions)
+
+    def _find_words(self, dictionary: Dictionary, position: List) -> List[str]:
```
jdb: Maybe you want to rename position to positions?
A more precise type is List[Tuple[int, int]]

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,126 @@
...
+        """ Method to return all words from given position """
+        words = []
```
jdb: DO you want the solution to mention the same words multiple times? I think you should use
words = set()

Below, you can replace words.extend() with word.update()

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,126 @@
...
+            return []
+        if (dictionary.is_word(word)):
+            words.append(word)
+
+        neighbours = self._get_neighbours(position[-1][0], position[-1][1])
+        for pos in neighbours:
+            if pos not in position:
+                position.append(pos)
+                words.extend(self._find_words(dictionary, position))
+                position.remove(pos)
```
jdb: position.pop() is O(1) and removes the last element. This is simpler and better complexity than .remove(element), because in this case, .remove has to go through the whole list to find the element to remove.

Remember, there is no magic, a list is not indexed like a dictionary, so even wihtout looking at the documented complexity, of the remove method, you can predict this method is O(n).

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,126 @@
...
+    def _get_word(self, positions: List) -> str:
+        """ Method to find corresponding word given a list of positions in grid """
+        return ''.join(self.grid[pos[0]][pos[1]] for pos in positions)
+
+    def _find_words(self, dictionary: Dictionary, position: List) -> List[str]:
+        """ Method to return all words from given position """
+        words = []
+
+        word = self._get_word(position)
+        if (not dictionary.is_prefix(word)):
```
jdb: the outer parens are not needed.

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,126 @@
...
+            return []
+        if (dictionary.is_word(word)):
```
jdb: same here.

sridha2412/assignment3/wordsearch.py
```diff
@@ -0,0 +1,126 @@
...
+
+        return words
+
+    def find_all_words(self, dictionary: Dictionary) -> List[str]:
+        """ Method to return List of all words found in grid """
+        words = []
+        for i in range(len(self.grid)):
+            for j in range(len(self.grid[i])):
+                for word in self._find_words(dictionary, [(i, j)]):
+                    if word not in words:
```
jdb: Be careful to use a set for this use case: the complexity of "in" on a list is problematic: for each word found, the whole list is checked.
Remember that sets are really cool, and although "in" is very readable, the complexity depends on the datastructure, and on the implementation of the __contains__ method.

jdb (changes requested):

sridha2412/assignment3/wordsearch.py
Sridha2412: thanks JD, I have pushed the new changes

jdb (approved):

PR #20, opened 2018-06-19
Epinar 3

evinpinar:
- There is a referenced "isVisited" array problem. How can we send a copy at each recursive call?
- Should we solve this by yielding? -> But then, how will we able to search "CARD" after finding the word "CAR"?

### DIFF
--- a/epinar/assignment3/word_search.py
+++ b/epinar/assignment3/word_search.py
@@ -1,0 +1,127 @@
+import string
+import unittest
+
+
+class Dictionary:
+
+	def __init__(self, words, prefixes):
+		"""
+		The dictionary class to keep the values.
+		:param words: Valid entries in a dictionary.
+		:type words: Array of strings.
+		:param prefixes: Valid prefixes of each words in the dictionary.
+		:type prefixes: Array of strings.
+		"""
+		self.words = words
+		self.prefixes = prefixes
+
+	def isWord(self, s):
+		"""
+		Returns whether the given string is a valid word.
+		:param s: Word query.
+		:type s: String
+		:return: If the string is valid.
+		:rtype: bool
+		"""
+		return s in self.words
+
+	def isPrefix(self, s):
+		"""
+		Returns whether the given string is a prefix of at least one word in
+		the dictionary.
+		:param s: Prefix query.
+		:type s: String
+		:return: If the string is valid.
+		:rtype: bool
+		"""
+		return s in self.prefixes
+
+
+def word_search(grid, dict):
+	"""
+	Given a grid of letters and a dictionary, it finds all the words from the
+	dictionary that can be formed in the grid.
+	:param grid: The grid of characters to search in.
+	:type grid: 2 dimensional array of characters, M x N .
+	:param dict: The dictionary of words and prefixes.
+	:type dict: Dictionary class object.
+	:return: Set of all words found in the grid.
+	:rtype: Array of strings
+	"""
+
+	found_words = []
+	new_words = []
+	M = len(grid)
+	N = len(grid[0])
+
+	for k in range(M):
+		for l in range(N):
+			visited_cache = [[False for _ in range(N)] for _ in range(M)]
+			print("Start looking... ", k, l)
+			new_words += _search(grid, dict, "", k, l, visited_cache, [])
+			found_words += list(set(new_words))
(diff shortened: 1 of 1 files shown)


epinar/assignment3/word_search.py
```diff
@@ -0,0 +1,128 @@
+import string
+import unittest
+
+
+class Dictionary:
+
+	def __init__(self, words, prefixes):
+		"""
+		The dictionary class to keep the values.
+		:param words: Valid entries in a dictionary.
```
jdb: can you use the Python type annotations instead of these syntax for the function signature?

https://docs.python.org/3/library/typing.html

epinar/assignment3/word_search.py
```diff
@@ -0,0 +1,128 @@
...
+	"""
+
+	found_words = []
+	new_words = []
+	M = len(grid)
+	N = len(grid[0])
+
+	for k in range(M):
+		for l in range(N):
+			visited_cache = [[False for _ in range(N)] for _ in range(M)]
```
jdb: visited_cache = [[False] * N] for _ in range(M)]
Yes, in Python, the list can be multiplied: the new list has the *same elements* repeated, many references to the same object instance.

See the emphasis on *same* above? If the element is immutable like False, it is ok to have many references toward it, but if the element is mutable, this won't work in ourcase, we do not want to have a list of references to the same list of False. A mutation through any reference modifies the list pointed to by all references

>>> [False] * 10
[False, False, False, False, False, False, False, False, False, False]

>>> l = [[1, 2]] * 2
[[1, 2], [1, 2]]

>>> l[0][0] = 9
[[9, 2], [9, 2]]

epinar/assignment3/word_search.py
```diff
@@ -0,0 +1,128 @@
...
+			print("Start looking... ", k, l)
+			new_words += _search(grid, dict, "", k, l, visited_cache, [])
```
jdb: found_words.update(_search(grid, dict, "", k, l, visited_cache, []))

jdb (changes requested):

epinar/assignment3/word_search.py
```diff
@@ -0,0 +1,128 @@
...
+		isVisited[i][j] = False
+		return found_words
+
+	if dict.isWord(s):
+		found_words.append(str(s))
+
+	s = s + grid[i][j]
+
+	if dict.isPrefix(s):
+		isVisited[i][j] = True
```
jdb: Try to reset isVisited to False before exiting the function so the slot becomes free to be visited from another neighbor.
The advantage is that you can pass the same mutable state to all recursive call, without copying isVisited at each function call.

jdb:
Please be more precise, more specific. What do you want to yield exactly?
One way to solve this problem with a generic, methodic way is to think about it as a graph traversal problem: a node is an (i, j) position in the grid, and line of ancestors (a "path") is a prefix.

Then you can solve the problem in 2 steps:
- build the right traversal function
- use the list of elements returned by the traversal to only keep the ones you want

When you traverse a graph, there is usually two things you may want to yield:
1. yield each node
2. yield each path of nodes

In this context, you want 2. and you want to build a function over the traversal that translate a path  into a prefix, and keep the prefix that are word.

If you can do this, then I explain how to implement the "graph pruning" with the heuristics that bad prefix will never becomes words.

jdb (changes requested):

epinar/assignment3/word_search.py
evinpinar: I am not sure if I understand your point. I think we can visit each node once, no? So, if we include a letter, it should stay visited. however, in the end, when we return to the search's starting point, we should reset the nodes that weren't included.

jdb: Can you add the following unit test?

words = ['crash', 'crush']
self.assertEqual(words, word_search(['crash', 'zzuzz'], words))

I think if you do not remove the current letter from Visited on the way up from the recursion, then in the example above, the  r for example is going to be visited, 'crash' will be found, but the r will not be re-used to create 'crush'. Both should be found.

jdb (changes requested):

evinpinar:
I added the extra unit test and now it works, thanks

jdb (approved):

PR #21, opened 2018-06-22
Silver solver 3

SilverSolver:


SilverSolver/assignment3/dictionary.py
```diff
@@ -0,0 +1,62 @@
...
+        if iterable:
+            for word in iterable:
+                self.add(word)
+    
+    def add(self, word: str):
+        node = self
+        for char in word:
+            found_in_child = False
+            # Search for the character in the children of the present `node`
+            for child in node.children:
```
jdb: This loop is in O(n). It uses a list of Dictionary, and look up the Dictionary with a special character.
Can speed up this look up by storing the Dictionaries in a Python dict instead of Python list?

SilverSolver/assignment3/dictionary.py
```diff
@@ -0,0 +1,62 @@
...
+        # This helper function returns tuple with meaning (isWord, isPrefix)
+        node = self
+        # If the self node has no children, then return False.
+        # Because it means we are trying to search in an empty trie
+        if not self.children:
+            return False, False
+        for char in prefix:
+            char_not_found = True
+            # Search through all the children of the present `node`
+            for child in node.children:
```
jdb: Same here, this is O(n) while the standard Python dict operates in O(1).

SilverSolver/assignment3/dictionary.py
```diff
@@ -0,0 +1,62 @@
...
+                    found_in_child = True
+                    break
+            # We did not find it so add a new chlid
+            if not found_in_child:
+                new_node = Dictionary(char)
+                node.children.append(new_node)
+                node = new_node
+        node.word_finished = True
+
+    def isPrefixOrWord(self, prefix: str) -> Tuple[bool, bool]:
```
jdb: Nicer API than what the exercice was recommending!!
I like it that you do not to have to call 2 methods on the Trie, that basically do the same thing .

SilverSolver/assignment3/word_search.py
```diff
@@ -0,0 +1,49 @@
+from bisect import bisect_left
+from dictionary import Dictionary
+from typing import List, Tuple
+
+def empty_matrix(m: int, n: int, value: bool) -> List[List[bool]]:
```
jdb: You can write this in one line: 
  return [[value]*n for _ in range(m)]

From there, maybe you just inline this expression, for concision, if you think it is clearer.

SilverSolver/assignment3/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+    for i in range(m):
+        result.append([value] * n)
+    return result
+
+def word_search(grid: List[List[str]], dictionary: Dictionary) -> List[str]: 
+    # -- Should we check grid correctness?
+    def neighbours(coords: Tuple[int]) -> Tuple[Tuple[int]]:
+        a, b = coords
+        neigh_coords = [(a + i, b + j) for i in [-1, 1, 0] for j in [-1, 1, 0]]
+        del neigh_coords[-1]
```
jdb: another way to do the same thing:
  [(a + i, b + j) for i in [-1, 1, 0] for j in [-1, 1, 0] if (i, j) != (0, 0)]
Up to you.

SilverSolver/assignment3/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+        return filter(lambda x: x[0] >= 0 and x[1] >= 0 and x[0] < m and x[1] < n, neigh_coords)
+    
+    def search(coords: Tuple[int]):
```
jdb: Unlike List[type] that takes one type and the elements should be matching the type, Tuple takes as many types as there are elements. Tuples are not really supposed to be of variable length, especially since they are immutable.

I think the correct type is Tuple[int, int]

SilverSolver/assignment3/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+
+    words = set()
+    if len(grid) > 0:
+        m, n = len(grid), len(grid[0])
+    else:
+        return set()
+    
+    history = empty_matrix(m, n, False) # used to track if grid point was visited to avoid cycles.
+    prefix = []                         # list of chars to store current prefix. 
+    # We use list instead of str because it allows adding and removing last element at cost of O(1).
```
jdb: Nice consideration. 
I have read and tested that appended to a string with += is in constant time. There is a somwhat recent optimization in the interpreter (maybe Python 2.6 or 2.7).

I agree with you that and I do not think popping the last element is O(1)

SilverSolver/assignment3/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+
+def empty_matrix(m: int, n: int, value: bool) -> List[List[bool]]:
+    # value should has immutable type!
+    result = []
+    for i in range(m):
+        result.append([value] * n)
+    return result
+
+def word_search(grid: List[List[str]], dictionary: Dictionary) -> List[str]: 
+    # -- Should we check grid correctness?
```
jdb: It is really good for the interviewer to get to know if the candidate thought about validation of the input. Some candidate would forget about it. Just clearly mention your assumption that the grid is correct, and that your priority is results, concision and clarity in the context of interviews. You can mention to the interviewer that you will implement that method iff there is time left.

SilverSolver/assignment3/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+    def neighbours(coords: Tuple[int]) -> Tuple[Tuple[int]]:
+        a, b = coords
+        neigh_coords = [(a + i, b + j) for i in [-1, 1, 0] for j in [-1, 1, 0]]
+        del neigh_coords[-1]
+        return filter(lambda x: x[0] >= 0 and x[1] >= 0 and x[0] < m and x[1] < n, neigh_coords)
```
jdb: 2 notes:
- it is recommended to use a list comprehension instead of creating a lambda. Last time I checked, it was also faster.
- Python supports interval checks as boolean expression: this is correct: 0 <= a < max
- Python supports unpacking. Ex: [(element[1], element[0]) for element in my_list] is the same as [(y, x) for x, y in my_list]

You can write this:
  return [(x, y) for x, y in neigh_coords if 0 <= 0 x < m and 0 <= y < m]

jdb (changes requested):

SilverSolver/assignment3/word_search.py
SilverSolver: How could they do such optimization if string is immutable? :open_mouth:

SilverSolver: List carries pointers to its beginning and end, that's why I thought that deleting or adding the last element is O(1)

SilverSolver/assignment3/word_search.py
SilverSolver: There are two misprints there. I didn't notice them first and didn't understand why tests failed :smiley: 
I think, you mean **[(x, y) for x, y in neigh_coords if 0 <= x < m and 0 <= y < n]** - this works correctly.

SilverSolver/assignment3/dictionary.py
SilverSolver: That's true! But I'm not sure that I need dict here necessarily. Set also allow lookup with O(1) time, that's why I decided to use it in the next variant of the solution (see last commit)

SilverSolver/assignment3/word_search.py
```diff
@@ -2,22 +2,15 @@
...
 def word_search(grid: List[List[str]], dictionary: Dictionary) -> List[str]: 
-    # -- Should we check grid correctness?
+    # From there, we suppose that grid is correct
     def neighbours(coords: Tuple[int]) -> Tuple[Tuple[int]]:
         a, b = coords
-        neigh_coords = [(a + i, b + j) for i in [-1, 1, 0] for j in [-1, 1, 0]]
-        del neigh_coords[-1]
-        return filter(lambda x: x[0] >= 0 and x[1] >= 0 and x[0] < m and x[1] < n, neigh_coords)
+        neigh_coords = [(a + i, b + j) for i in [-1, 1, 0] for j in [-1, 1, 0] if (i, j) != (0, 0)]
+        #return filter(lambda x: x[0] >= 0 and x[1] >= 0 and x[0] < m and x[1] < n, neigh_coords)
```
jdb: Please remove "dead code".

SilverSolver/assignment3/dictionary.py
jdb: It is true that the set is O(1) for the presence check, but here you do not use the presence check, you use the for loop over all the elements in the set. 
If you do a for loop over many, more than one element, then that algorithm is automatically more than O(1).

# This code is O(n) with a set
a_set_of_2_tuples = {('a', 1), ('b', 2), ('c', 3), ('d', 4)}
found = None
for (key, value) in a_set_of_2_tuples:
  if key == 'c':
    found = value

# This code is O(1) with a dict
a_dict_of_2_tuples = {'a:, 1, 'b': 2, 'c': 3, 'd': 4}
found = a_dict_of_2_tuples['c']

If you need to find an Trie Node by its 'char' attribute, then you need to 'key' a dict of Trie Nodes by the char...

jdb (changes requested):

SilverSolver/assignment3/dictionary.py
```diff
@@ -0,0 +1,40 @@
+from typing import Tuple
+
+class Dictionary:
+    """
+    I used some code from here to implement the trie -
```
jdb: maybe you can remove this mention if your code and the original code diverged enough.

jdb (approved):

jdb (approved):
Great thanks!

PR #22, opened 2018-06-22
A cool solution with user-controlled graph traversal pruning for assignment 3

jdb:


evinpinar:
It's a great solution, thanks!

SilverSolver:
I'm not sure - where exactly user controls graph traversal pruning? :smiley:

jdb:
Ok, "pruning" is cutting a whole area of the graph and we do not even look into it. The term is used in gardening as well, for real tree branches.

jdb:
And in this solution, we have 2 parts: the traversal function that generates all the path and the functions that uses it to filter the path.

jdb:
A traversal function that be communicated to stop recursing from the outside (from the caller) is what I find interesting

jdb:
"Pruning a graph traversal can be controlled from the caller" if this is a bit easier to understand.

jdb:
I hope I am not confusing you too much with yield and all that, maybe I will send you a solution without yield at all.

SilverSolver:
No problem! I think I've understand the "yield" part here.
Thanks for explanation, it is more clear now what did you mean

jdb/assignment3/wordsearch.py
```diff
@@ -0,0 +1,138 @@
...
+    grid = Grid('abc', 'def', 'ghi', lexicon=True)
+    for letter, position in zip('abcdefghi',
+                                ((x, y) for y in range(3) for x in range(3))):
+      self.assertEqual(grid[position], letter)
+
+  def test_words(self):
+    words = Grid('aar', 'tcd').words()
+    for word in ['card', 'data', 'act', 'arc', 'cad', 'car', 'cat', 'rat',
+                 'rca', 'tad', 'tar', 'ac', 'ad']:
+      self.assertIn(word, words)
```
SilverSolver, line 134: :slightly_frowning_face: Unittest give the following error:

**FAIL: test_words (wordsearch.TestGrid)
  File "/mnt/FileStorage/Projects_progs/ForVacancies/CodeU-2018-Group2/jdb/assignment3/wordsearch.py", line 138, in test_words
    self.assertIn(word, words)
AssertionError: 'data' not found in {'tad', 'cat', 'tc', 'tar', 'cr', 'act', 'rd', 'rca', 'card', 'ta', 'ar', 'car', 'arc', 'cd', 'cad', 'ad', 'ca', 'at', 'ac'}**

Then if I'm deleting the 'data', it shows the same error for 'rat'. The rest of words are found correctly.
What's wrong? :thinking: Both words 'data' and 'rat' contains diagonal step, but 'rca', for example, contains it too and not causing the error.

SilverSolver (changes requested):
The last test cause an error

SilverSolver (approved):

PR #23, opened 2018-07-01
sridha2412-4 q1

Sridha2412:


sridha2412/assignment4/counting_islands.py
```diff
@@ -0,0 +1,48 @@
...
+import unittest
+
+class Grid:
+    def __init__(self, grid: List):
+        """ Constructor to initialize a grid """
+        self.grid = grid
+        self.row = len(self.grid)
+        self.col = len(self.grid[0])
+
+    def _get_neighbours(self, x: int, y: int) -> List[Tuple[int, int]]:
```
jdb: To me it looks like get_neighbours could be simplified by merging it with  find_land. Another approach is to merge it  with __init__ and make a set of of all unvisited *land* position. Unvisited can then shrink as you explore each islands.

jdb:
Excellent! This is clear and simple code. Thanks for making it clear between private and public functions: it was clear to me which were the important functions and which were the helpers.

2 things you can think of:

1. set a docstring to your module that summarizes in plain English your reasoning.

2. have you thought about a simpler function instead of a class. Functions are simpler to read, they usually have one starting point and one exit points. With object oriented programming, it is a bit more complicated which functions is a callback and gets automatically called, and which functions are called by others...

jdb (changes requested):
Have you thought about turning the input grid into a set of all unvisited position that are on the land?

This way you would not need to keep both visited and the input grid.

Sridha2412:
@jdb not sure if this is what you meant by not having to keep both visited and input grid but I am modifying the original grid

jdb (approved):
I will send my own solution later today I believe. There, I did not modify the input but used a shrinking set of unvisited position.

PR #24, opened 2018-07-02
Silver solver 4

SilverSolver:
4 assignment
(remark: I already solved this problem at pramp.com and used code (with some corrections and additions) and tests from it)

### DIFF
--- a/SilverSolver/assignment4/islands.py
+++ b/SilverSolver/assignment4/islands.py
@@ -1,0 +1,44 @@
+from collections import deque
+from typing import List
+
+def neighbourhood(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
+    # helper function, returns the neighbourhood of (i, j) cell in islandMatrix
+    neigh = []
+    if i > 0:
+        if islandMatrix[i - 1][j] == 1:
+            neigh.append((i - 1, j))
+    if i < len(islandMatrix) - 1:
+        if islandMatrix[i + 1][j] == 1:
+            neigh.append((i + 1, j))
+    if j > 0:
+        if islandMatrix[i][j - 1] == 1:
+            neigh.append((i, j - 1))
+    if j < len(islandMatrix[0]) - 1:
+        if islandMatrix[i][j + 1] == 1:
+            neigh.append((i, j + 1))
+    return neigh
+
+def island_fill(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
+    # helper function, fills the island, to which (i, j) belongs, with "2"
+    # it will help to distinct it with the islands in which we've never been
+    q = deque()
+    q.append((i, j))
+    while q:
+        element = q.pop()
+        if islandMatrix[element[0]][element[1]] == 1:
+            for e in neighbourhood(element[0], element[1], islandMatrix):
+                q.append(e)
+            islandMatrix[element[0]][element[1]] = 2
+    return islandMatrix
+
+def get_number_of_islands(islandMatrix: List[List[int]]) -> int:
+    # the essential function
+    islandNum = 0
+    for i in range(len(islandMatrix)):
+        for j in range(len(islandMatrix[0])):
+            if islandMatrix[i][j] != 1:
+                pass
+            else:
+                islandNum += 1
+                islandMatrix = island_fill(i, j, islandMatrix)
+    return islandNum
--- a/SilverSolver/assignment4/islands.py
+++ b/SilverSolver/assignment4/islands.py
@@ -1,28 +1,20 @@
-from collections import deque
+from copy import deepcopy
 from typing import List
 
 def neighbourhood(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
-    # helper function, returns the neighbourhood of (i, j) cell in islandMatrix
-    neigh = []
-    if i > 0:
-        if islandMatrix[i - 1][j] == 1:
-            neigh.append((i - 1, j))
-    if i < len(islandMatrix) - 1:
-        if islandMatrix[i + 1][j] == 1:
-            neigh.append((i + 1, j))
-    if j > 0:
-        if islandMatrix[i][j - 1] == 1:
-            neigh.append((i, j - 1))
-    if j < len(islandMatrix[0]) - 1:
-        if islandMatrix[i][j + 1] == 1:
-            neigh.append((i, j + 1))
-    return neigh
+    "helper function, returns the neighbourhood of (i, j) cell in islandMatrix"
+
+    for di, dj in ((-1, 0), (1, 0), (0, -1), (0, 1)):
+        ni, nj = di + i, dj + j # ni, nj are a neighbor candidates for i, j
+        if 0 <= ni < len(islandMatrix) and 0 <= nj < len(islandMatrix[0]) and islandMatrix[ni][nj]:
+            yield ni, nj
 
 def island_fill(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
-    # helper function, fills the island, to which (i, j) belongs, with "2"
-    # it will help to distinct it with the islands in which we've never been
-    q = deque()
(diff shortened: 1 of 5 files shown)


SilverSolver/assignment4/islands.py
```diff
@@ -0,0 +1,44 @@
+from collections import deque
+from typing import List
+
+def neighbourhood(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
+    # helper function, returns the neighbourhood of (i, j) cell in islandMatrix
```
jdb: nit: turn the comment into a Python docstring:
"helper function, returns the neighbourhood of (i, j) cell in islandMatrix."

A Python docstring is technically the first line after the declaration and it must be a literal string.

SilverSolver/assignment4/islands.py
```diff
@@ -0,0 +1,44 @@
...
+            neigh.append((i, j - 1))
+    if j < len(islandMatrix[0]) - 1:
+        if islandMatrix[i][j + 1] == 1:
+            neigh.append((i, j + 1))
+    return neigh
+
+def island_fill(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
+    # helper function, fills the island, to which (i, j) belongs, with "2"
+    # it will help to distinct it with the islands in which we've never been
+    q = deque()
```
jdb: The way you use it, it looks like a simple list should work. Did I miss something?

SilverSolver/assignment4/islands.py
```diff
@@ -0,0 +1,44 @@
+from collections import deque
+from typing import List
+
+def neighbourhood(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
+    # helper function, returns the neighbourhood of (i, j) cell in islandMatrix
+    neigh = []
```
jdb: for di, dj in ((-1, 0), (1, 0), (0, -1), (0, 1)):  # di, dj for deltai deltaj
  ni, nj = di + i, dj + j  # ni, nj is a neighbor candidate for i, j
  if 0 <= ni < len(islandMatrix) and 0 <= nj < len(islandMatrix[0]) and islandMatrix[ni][nj]:
    yield ni, nj

I find this version of neighbor a bit easier to read: 
- the neighbor candidate are more easily identified at the top of the for loop,
- the guards that decide whether a candidate is good or not are clearly shown after an 'if' clauses
- we avoid the repetition of the same pattern of code which could have a slight typos
- nit: no need to create and empty list, yield saves two lines, and is slightly more efficient

SilverSolver/assignment4/islands.py
```diff
@@ -0,0 +1,44 @@
...
+    # helper function, fills the island, to which (i, j) belongs, with "2"
+    # it will help to distinct it with the islands in which we've never been
+    q = deque()
+    q.append((i, j))
+    while q:
+        element = q.pop()
+        if islandMatrix[element[0]][element[1]] == 1:
+            for e in neighbourhood(element[0], element[1], islandMatrix):
+                q.append(e)
+            islandMatrix[element[0]][element[1]] = 2
```
jdb, line 23: I think it would be clearer to keep a set of visited position, it is quite a surprise that the function that counts the island actually mutates the input. This means that the user who needs to use your function would need to make a copy before using your function if they need to re-use their input somewhere else.

SilverSolver/assignment4/islands.py
```diff
@@ -0,0 +1,44 @@
...
+    return islandMatrix
+
+def get_number_of_islands(islandMatrix: List[List[int]]) -> int:
+    # the essential function
+    islandNum = 0
```
jdb: Have you thought about creating a easier datastructure that the matrix of 0 and 1? I suggest you create a simple set of **unvisited** positions that are **on the land**. As you visit the land position, you can pop the position from the island. You can actually discard all "lake" position very early in the algorithm and not be concerned about them anymore.

Like in the previous exercice, only the function that receives the input matrix and the function generating the neighbors really have to know about the implementation of a position, the rest of the algorithm does not have to know this implementation detail.

jdb (changes requested):

SilverSolver/assignment4/islands.py
SilverSolver: I think you are right because I use pop from only one side of a data structure.

SilverSolver/assignment4/islands.py
```diff
@@ -0,0 +1,36 @@
...
+        ni, nj = di + i, dj + j # ni, nj are a neighbor candidates for i, j
+        if 0 <= ni < len(islandMatrix) and 0 <= nj < len(islandMatrix[0]) and islandMatrix[ni][nj]:
+            yield ni, nj
+
+def island_fill(i: int, j: int, islandMatrix: List[List[int]]) -> List[List[int]]:
+    """
+    helper function, fills the island, to which (i, j) belongs, with "2"
+    it will help to distinct it with the islands in which we've never been
+    """
+    q = [(i, j)]
```
jdb, line 17: Please rename q to something more meaningful, or write in the docstring the pseudo code of the algorithm :)

Do you mean 'queue'? Does this data structure holds the role of a 'visited' datastructure?

+    q = [(i, j)]
￼ +    while q:
￼ +        ei, ej = q.pop()
￼ +        if islandMatrix[ei][ej] == 1:
￼ +            for e in neighbourhood(ei, ej, islandMatrix):
￼ +                q.append(e)
￼ +            islandMatrix[ei][ej] = 2

One way to simplify this algorithm, IMHO is use the same datasctructure for q and islandMatrix. They mean almost the same thing: when q is empty, you have visited all the current island, and when islandMatrix has all the 1 turned into 2, you have also visited all the islands.

SilverSolver/assignment4/tests.py
```diff
@@ -0,0 +1,33 @@
...
+        self.assertEqual(get_number_of_islands([[1]]), 1)
+
+    def test_1(self):
+        self.assertEqual(get_number_of_islands([[1,0,1,0]]), 2)
+
+    def test_2(self):
+        self.assertEqual(get_number_of_islands([[1,0,1,0],[0,1,1,1],[0,0,1,0]]), 2)
+        self.assertEqual(get_number_of_islands( \
+                [[1,0,1,0],[0,1,1,1],[0,0,1,0],[1,1,0,0],[0,1,0,1]]), 4)
+        self.assertEqual(get_number_of_islands( \
```
jdb, line 27: The slash is not needed. The Python parser is flexible enough to note that the function call continues on the next line.

SilverSolver/assignment4/tests.py
```diff
@@ -0,0 +1,33 @@
...
+        self.assertEqual(get_number_of_islands([[0]]), 0)
+        self.assertEqual(get_number_of_islands([[1]]), 1)
+
+    def test_1(self):
+        self.assertEqual(get_number_of_islands([[1,0,1,0]]), 2)
+
+    def test_2(self):
+        self.assertEqual(get_number_of_islands([[1,0,1,0],[0,1,1,1],[0,0,1,0]]), 2)
+        self.assertEqual(get_number_of_islands( \
+                [[1,0,1,0],[0,1,1,1],[0,0,1,0],[1,1,0,0],[0,1,0,1]]), 4)
```
jdb, line 26: Please align the examples vertically, so that we can vizualize the islands better :)
You can also use strings that get_number_of_islands transforms into a grid.
This is also equivalent to the deepcopy you currently use in terms of performance.

self.assertEqual(
    get_number_of_islands(
        '01010',
        '00111',
        '10010',
        '01100',
        '10101'), 6)

Well, I think this example show that the algorithm is incorrect: I count 8 islands here. 

I think a small counter example is:

self.assertEqual(
    get_number_of_islands(
        '01',
        '01'), 1)

if this works please add the unit test:

self.assertEqual(
    get_number_of_islands(
        '010',
        '010',
        '010'), 1)

jdb (changes requested):

jdb (approved):

PR #25, opened 2018-07-02
Crossopt 4

crossopt:


crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+
+
+# disjoint set union for grid components
```
jdb: Should this comment be a docstring?
Here are general guidelines about comments:
- docstrings: public documentation for the object
- comments starting with #: explanation to the reader of the code about the details of the implementation.

crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,53 @@
...
+class _GridDSU:
+    def __init__(self, n, m):
+        self.parent = dict()
+        self.size = dict()
+        for i in range(n):
+            for j in range(m):
+                self.parent[(i, j)] = (i, j)
```
jdb: The parentheses on both lines are unneeded. Note: the comma is the bit of syntax that is the constructor for the type "tuple". 
The parentheses in Python are here for the correction interpretation of priority.

This said, can you find the exception to this rule above, that the constructor of tuple is the comma?

crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,53 @@
...
+        self.parent[v] = u
+
+
+class Grid(object):
+    def __init__(self, n, m, grid):
+        self.n = n
+        self.m = m
+        self.grid = grid
+
+    # returns DSU for grid where all islands are components
```
jdb: What do you call components? 
Can you define this for your poor reviewer :)

crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,53 @@
...
+        if self.parent[u] != u:
+            self.parent[u] = self[self.parent[u]]
+        return self.parent[u]
+
+    def join(self, u, v):
+        u = self[u]
+        v = self[v]
+        if self.size[u] < self.size[v]:
+            u, v = v, u
+        self.size[u] += self.size[v]
```
jdb: I think you need to also set the size for size[v] to be equal to size[u], no?

crossopt/assignment4/countislands.py
jdb: Can you put a link for what is a disjoint set union?

crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Grid(object):
+    def __init__(self, n, m, grid):
+        self.n = n
+        self.m = m
+        self.grid = grid
+
+    # returns DSU for grid where all islands are components
+    def colorIslands(self):
```
jdb: Can you explain what you mean by "color" and why colors is somehow equivalent to DSU?
Do you think you can have a clearer name like MakeDsu?

Currently, there are 2 classes and 6 methods in total: and it is non-obvious what are the protocol or the contract between the two. Can you simplify this?

For example: if this method just returns the Dsu, then the feature might better placed on the __init__ of _GridDSU?
Currently Grid must know the _GridDSU constructor, the join method and the getter, it looks like if this method is moved to the _GridDSU constructor, then Grid only needs to know the _GridDSU constructor.

jdb (changes requested):
Good job for finding a solution! Maybe now try to simplify it, 2 classes and 3 methods each is complex, i think we would all be confused during the interview.

I suggest you cram everything  into one long funciton and see if there is room for simplification. One function for all this is maybe a bit radical but you might find a good balance.

crossopt/assignment4/countislands.py
crossopt: Ok, my code for this assignment was really awkward to look at several days later. And I still think that when I see it next time I'll be able to simplify it again:)

I don't need to set size[v], because in my implementation of the DSU, I only ever need to know the correct size of a vertice that is the root of its tree, and v is joined to u and as such is not a root

crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,43 @@
...
+        for i in range(n):
+            for j in range(m):
+                self.parents[i, j] = (i, j)
+                self.sizes[i, j] = 1
+                if i and self.grid[i][j] and self.grid[i - 1][j]:
+                    self.join((i, j), ((i - 1, j)))
+                if j and self.grid[i][j] and self.grid[i][j - 1]:
+                    self.join((i, j), ((i, j - 1)))
+
+    # returns unique id for each island
```
jdb: Please use docstrings :)

crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,43 @@
...
+    def __getitem__(self, u):
```
jdb: I do not understand what is the role of the getitem. Can you document it?
I found it surprising that a call to __getitem__ mutates the internal state.

jdb: Specifically, I do not understand when this important step is designed to happen: 
1. at construction time? 
2. or at countIsland time?

As I read it, the mutation can happen at both times.

jdb: Oh, I now see that __getitem__ is recursive, on top of mutating. This confuses the complexity. 

1. Can you rename __getitem__ to something more meaningful?
2. Can you make sure this method is called only at __init__ or at countIsland time, but not both?

crossopt/assignment4/countislands.py
```diff
@@ -0,0 +1,43 @@
...
+        if self.parents[u] != u:
+            self.parents[u] = self[self.parents[u]]
+        return self.parents[u]
+    
+    def join(self, u, v):
```
jdb: What is the complexity of join()?
Due to the recursive nature of __getitem__ it looks like it is O(n). 
I think this makes the  n^2. Am I right?

jdb (changes requested):

jdb (approved):

crossopt/assignment4/countislands.py
crossopt: With both path compression and union by size the complexity of join and get in DSU is O(alpha(n))

crossopt/assignment4/countislands.py
crossopt: getitem returns a single unique item for all of the elements in a set.
The mutation of the parents is just an internal part of the operation that improves the complexity

PR #26, opened 2018-07-02
Epinar 4

evinpinar:


epinar/assignment3/word_search.py
```diff
@@ -0,0 +1,92 @@
+import string
```
jdb, line 1: Can you remove this file from the pull request?

epinar/assignment4/islands.py
```diff
@@ -0,0 +1,56 @@
+import unittest
+
+
+def visit_island(grid, k, l, M, N):
+	""" Marks the visited island pieces to not to visit in the future again."""
+	if 0 <= k < M and 0 <= l < N:
+		if grid[k][l] is True:
```
jdb, line 7: nit: simply 

if grid[k][l]:

"if grid[k][l]" and "if grid[k][l] is True" are equivalent and the latter is more verbose.

epinar/assignment4/islands.py
```diff
@@ -0,0 +1,56 @@
...
+			visit_island(grid, k + 1, l, M, N)
+			visit_island(grid, k - 1, l, M, N)
+			visit_island(grid, k, l - 1, M, N)
+			visit_island(grid, k, l + 1, M, N)
+	return
+
+
+def count_islands(grid):
+	""" Counts the number of islands """
+	visited = grid.copy()  # copy the grid in order not to lose the real information.
```
jdb, line 18: I suggest you have two datastructure:
- the input can be a list of strings: it is easier to read and make test
- the internal data structure can be simpler: you can start with a set of all invisited land position, and remove from the set the position you visit. This would also remove the confusion with grid and visited.

The rule, here, is: make sure to find the best datastructure for the problem at hand: this really simplifies your algorithm later.

jdb (changes requested):

jdb (approved):

PR #27, opened 2018-07-14
partial solution to question

Sridha2412:


group/assignment5/sridha2412.py
```diff
@@ -0,0 +1,71 @@
...
+            break
+    return i
+
+
+def create_DAG(words: List[str]) -> dict:
+    """ A method to create a direct acyclic graph from a list of words
+    Args:
+        words: list of words e.g ['ART', 'RAT', 'CAT', 'CAR']
+    Returns:
+        dict: a DAG e.g. {'A': [R], 'R': ['C'], 'C', 'T': ['R']}
```
epinar: Hey, I'm not sure why we have 'C' : 'R' connection here.

group/assignment5/sridha2412.py
```diff
@@ -0,0 +1,71 @@
...
+
+    alphabets = set()
+    for word in words:
+        alphabets = alphabets.union(word)
+
+    graph = {alpha: {"set": set(), "visited": False} for alpha in alphabets}
+
+    for word1, word2 in zip(words, words[1:]):
+        cp = common_prefix(word1, word2)
+        if(word1[cp] != word2[cp]):
```
epinar: Do we need to check it again?

group/assignment5/sridha2412.py
Sridha2412: Hey I have 'RAT' before 'CAT' so R comes before C and that is why I have 'R': ['C']. Is that what you mean?

group/assignment5/sridha2412.py
Sridha2412: Oh yeah, you're right, I didn't realise that. Thanks!

Sridha2412: Oh wait, no. In case I have a word like 'CAR' and 'CART', it would return cp = 2 so I would need the check

group/assignment5/sridha2412.py
```diff
@@ -0,0 +1,71 @@
...
+    Args:
+        word1: first word
+        word2: second word
+    Returns:
+        int: position end of common prefix
+    """
+
+    for i in range(0, min(len(word1), len(word2))):
+        if word1[i] != word2[i]:
+            break
```
evinpinar, line 14:
```
    if word1[i] != word2[i]:
        return i

return Null
```

and in the create_dag function:
```
cp = common_prefix(word1, word2)
if cp:
```

This might look better in order not to check the lengths again.

group/assignment5/sridha2412.py
evinpinar, line 23: Hey, I'm not sure why we have 'C' : 'R' connection here.

group/assignment5/sridha2412.py
```diff
@@ -0,0 +1,71 @@
...
+    return graph
+
+def _topological_sort(alpha: str, alphabets: List[str], graph: dict):
+    """ A recursive method to topologically sort a graph
+    Args:
+        alpha: character to check all outgoing nodes for
+        alphabets: list to store order of alphabets
+        graph: a DAG with all the graph vertices
+    """
+    graph[alpha]["visited"] = True
```
evinpinar, line 46: Follow up 2: We should raise error if there is any loop, which we can check here: ` if graph[alpha]["visited"]`. We might simply return the alpha, and a traceback character --which needs to be passed with the function as well-- to show the reason. 
Follow up 3: Instead of returning an error, we can break the search and continue by the next node.

group/assignment5/sridha2412.py
```diff
@@ -0,0 +1,71 @@
...
+        words: list of words e.g ['ART', 'RAT', 'CAT', 'CAR']
+    Returns:
+        dict: a DAG e.g. {'A': [R], 'R': ['C'], 'C', 'T': ['R']}
+    """
+
+    alphabets = set()
+    for word in words:
+        alphabets = alphabets.union(word)
+
+    graph = {alpha: {"set": set(), "visited": False} for alpha in alphabets}
```
evinpinar, line 30: Follow up 1: For detecting other possible topological sorts, we can add an "in degree" for each node, which will help us checking the same level nodes and proceed.

group/assignment5/sridha2412.py
SilverSolver: I agree with this

group/assignment5/sridha2412.py
```diff
@@ -0,0 +1,71 @@
+from typing import List, Tuple
+
+def common_prefix(word1: str, word2: str) -> int:
+    """ A method to find the common prefix in two words and return the end position
```
SilverSolver, line 4: Nice comments! They are helpful

group/assignment5/sridha2412.py
crossopt: Only it won't work if the common prefix length is 0, as in ART and RAT
It'll say that cp is 0 and won't go into that if, but we want to add an edge between A and R

I do agree that the extra length check is confusing, though.

group/assignment5/sridha2412.py
crossopt, line 30: Is whether a node is visited really a part of the graph structure?

crossopt: Oh, yeah, I suppose it could be.

Your solution might not need multiple dfs-like function that need a visited counter even with the optionals, so I suppose it doesn't matter either way

crossopt: I do want to complain about the set named "set", though:)

edges, adjacent, whatever, there probably is some more fitting name

PR #29, opened 2018-07-16
Crossopt 5

crossopt:


group/assignment5/find_alphabet.py
```diff
@@ -0,0 +1,114 @@
...
+    order = list()
+
+    for v in graph.keys():
+        if v not in used:
+            _topsort(v, graph, used, order)
+
+    return order[::-1]
+
+
+def fix_incorrect(words):
```
SilverSolver, line 109: This function should be very slow... we should think how to make it faster

group/assignment5/find_alphabet.py
```diff
@@ -0,0 +1,114 @@
...
+def _topsort(v, graph, used, order):
+    """ computes topsort of acyclic graph """
+    used.add(v)
+    for u in graph[v]:
+        if u not in used:
+            _topsort(u, graph, used, order)
+    order.append(v)
+
+
+def _find_cycle(graph, start):
```
SilverSolver, line 53: What is the purpose of this function? How finding a shortest cycle will help us to solve the problem?

group/assignment5/unittests.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+
+import unittest
+from find_alphabet import is_correct, find_one_alphabet, find_incorrect, fix_incorrect
+
+
+class TestDictionary(unittest.TestCase):
```
SilverSolver, line 7: Nice tests!

group/assignment5/find_alphabet.py
crossopt: Ok, I think this is me being confused about the assignment.

I read "minimal set of constraints on the alphabet" as "least amount of letters that can't be ordered", which is basically "shortest cycle".

Whereas it probably was "set of constraints such that removing any of them makes the problem solvable".

Ouch.

group/assignment5/find_alphabet.py
crossopt: Same here, probably just me weirdly interpreting the assignment.

"maximal (with as many words as possible) subset of the dictionary".

Actually though, on my second read-through I still think that means that we should leave the overall max possible amount of words.

In that case, the problem is actually NP-hard (https://en.wikipedia.org/wiki/Feedback_arc_set), so it's unlikely we'll be able to think of anything substantially faster:)

group/assignment5/find_alphabet.py
```diff
@@ -0,0 +1,114 @@
...
+            return True
+        elif not color[u] and _is_cycle(u, graph, color):
+            return True
+    color[v] = 2
+    return False
+
+
+def _topsort(v, graph, used, order):
+    """ computes topsort of acyclic graph """
+    used.add(v)
```
evinpinar, line 46: I think we should try to find a way to find maximal and minimal constraints during topsort -> how can we detect the cycle here?

group/assignment5/unittests.py
evinpinar: Indeed !

group/assignment5/find_alphabet.py
evinpinar: Yeah we should try out different combinations indeed, but maybe instead of different word/dictionary combinations, we can try to build up paths and remove the nodes etc... ?

randomUsernameAndPassword: Yes, it looks like we can't really improve it asymptotically.
@epinar: Do you have any more specific idea about how to implement removing only words from cycles?
Currently, graph has an information about the "successors" of a given letter, but we don't have any info about words representing this dependency. We could have this relation in a graph, but many different pairs of words could represent it, so it would be a huge drop of memory-efficiency.

randomUsernameAndPassword (commented):
I would just fix that part with minimal set of constraints.

PR #30, opened 2018-07-18
Epinar 5

evinpinar:
An alternative representation for graph, and the first followup which I solved by holding indegree of each node, and traverse all possible paths.


epinar/assignment5/unit_tests
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+
+import unittest
+from unknown_alphabet import find_alphabets, find_alphabet
+
+class TestDictionary(unittest.TestCase):
+
+	def check_equal(self, L1, L2):
+		return sorted(L1) == sorted(L2)
```
SilverSolver, line 9: I love this function. It is very convenient and I think we should use it in the tests for the final solution.

epinar/assignment5/unit_tests
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+
```
SilverSolver, line 2: It should be better to name this file "unit_tests.py" or like this
Lack of the extension is confusing

epinar/assignment5/unit_tests
```diff
@@ -0,0 +1,21 @@
...
+	def test_find_alphabets(self):
+		self.check_equal(find_alphabets(['ART', 'RAT', 'CAT']), [['T', 'A', 'R', 'C'], ['A', 'T', 'R', 'C'], ['A', 'R', 'T', 'C'], ['A', 'R', 'C', 'T']])
+		self.check_equal(find_alphabets(['ART', 'RAT', 'CAT', 'CAR']), [['T', 'A', 'R', 'C'], ['A', 'T', 'R', 'C']])
+		self.check_equal(find_alphabets(['S', 'ART', 'RAT', 'CAT', 'CAR']), [['S', 'A', 'T', 'R', 'C'], ['S', 'T', 'A', 'R', 'C'], ['T', 'S', 'A', 'R', 'C']])
+
+	def test_find_alphabet(self):
+		self.assertEqual(find_alphabet(["1a", "1b", "a1", "ab"]), ["1", "a", "b"])
+
+
```
SilverSolver, line 19: I modified my own tests a bit and added some new for testing your solution. Your solution passed these tests.
You can add them if you want:
```
    def test_0(self):
        self.assertEqual(find_alphabet([]), [])
        self.assertEqual(find_alphabets([]), [[]])

    def test_1(self):
        self.assertEqual(find_alphabets(["a"]), [["a"]])
        self.assertEqual(find_alphabet(["1", "2", "3"]), ["1", "2", "3"])
        self.check_equal(find_alphabets(["12"]), [["1", "2"], ["2", "1"]])

    def test_2(self):
        self.assertEqual(find_alphabets(["1a", "1b", "a1", "ab"]), [["1", "a", "b"]])
        self.check_equal(find_alphabets(["1b", "a1"]), [["1", "a", "b"], ["1", "b", "a"]])

    def test_fault(self):
        self.assertEqual(find_alphabets(["1", "2", "1"]), [[]])
        self.assertEqual(find_alphabets(["12", "22", "21"]), [[]])
```

epinar/assignment5/unknown_alphabet.py
```diff
@@ -0,0 +1,98 @@
...
+			cp = _common_prefix(word1, word2)
+			if cp != -1:
+				self.nodes[word1[cp]].add_edge(self.nodes[word2[cp]])
+
+	def __iter__(self):
+		for node in self.nodes.values():
+			yield node
+
+
+def _topsort1(v, graph, alphabet):
```
SilverSolver, line 35: I'm not sure that we need this function if we have a "_topsort2"

epinar/assignment5/unknown_alphabet.py
```diff
@@ -0,0 +1,98 @@
...
+	""" computes topsort of acyclic graph """
+	v.is_visited = True
+	for u in v.edges:
+		if not u.is_visited:
+			_topsort1(u, graph, alphabet)
+	alphabet.append(v.value)
+
+
+def find_alphabet(words):
```
SilverSolver, line 44: I'm not sure that we need this function if we have "find_alphabets"

epinar/assignment5/unknown_alphabet.py
```diff
@@ -0,0 +1,98 @@
...
+
+			v.is_visited = False
+			curr_alphabet.pop()
+			for u in v.edges:
+				u.indegree += 1
+
+			isDone = True
+
+	if not isDone:
+		alphabets.append(curr_alphabet.copy())
```
SilverSolver, line 85: As I can see, this function returns [[]] if we have an inconsistent dictionary. I really don't know is it ok or we should return something more special in this case...

SilverSolver (commented):
This code solves the problem, but is surplus somewhere

epinar/assignment5/unit_tests
evinpinar: Oops i lost the extension while moving files to manage the commit issues, will fix it 👍

epinar/assignment5/unit_tests
evinpinar: Definitely helpful!

epinar/assignment5/unit_tests
evinpinar: Oh perfect, thanks! In the future, after coming up with the main group branch, we should iterate on the code collaboratively :)

epinar/assignment5/unknown_alphabet.py
evinpinar: It was my attempt to solve the first followup question, to find multiple alphabets. 
For the second followup, ie. if the dictionary is inconsistent, the function should raise warning and show what creates inconsistency -> cycle. But for words: ['A', 'B', 'C', 'AB', 'AA'], it currently returns ['A', 'B', 'C']. It should return ['A'] or ['B'] or both. 
And for last followup, to find the maximal consistent set, I think we should find the minimum number of elements that will make the set consistent. For same example, removing the letter 'A' will help and the resulting alphabet will be: ['B', 'C'].

epinar/assignment5/unknown_alphabet.py
evinpinar: removed 👍

epinar/assignment5/unknown_alphabet.py
evinpinar: Removed 👍

crossopt: Ehh I'm confused. Why wouldn't we need this function?
After all, it's faster than find_alphabets

epinar/assignment5/unknown_alphabet.py
crossopt: Same here. Though the names are confusing, maybe something that actually shows what the second fancy topsort does?

PR #31, opened 2018-07-25
Create move_cars.py

crossopt:
Soo I think we'll have a lot to talk about with this assignment and there definitely will be different approaches, unlike the topsort.

All the better to start early!

### DIFF
--- a/group/assignment6/move_cars.py
+++ b/group/assignment6/move_cars.py
@@ -1,7 +1,7 @@
 def car_positions(state):
     """ given state returns array where ith element is the position of ith car """
-    positions = [0] * len(start)
-    for place, car in enumerate(start):
+    positions = [0] * len(state)
+    for place, car in enumerate(state):
         positions[car] = place
     return positions
 
--- a/group/assignment6/move_cars.py
+++ b/group/assignment6/move_cars.py
@@ -1,31 +1,80 @@
-def car_positions(state):
-    """ given state returns array where ith element is the position of ith car """
-    positions = [0] * len(state)
-    for place, car in enumerate(state):
-        positions[car] = place
-    return positions
+from typing import List
 
 
-def generate_moves(start, final):
-    """ ith element of start state is the car currently in the ith space
-    ith element of final state is the car that should be in ith space
-    empty space is 0
-    yields optimal sequence of moves in format (car number, old space, new space) """
+def check_input(start: List[int], final: List[int]) -> bool:
+    """ Method to check correctness of input
+    Args:
+        start: list of integers where ith element is the car currently in the ith space
+        final: list of integers where ith element is the car that should be in the ith space
+    Returns:
+        bool: is the input correct
+    """
+    start_set = set(start)
+    final_set = set(final)
+    car_set = set([i for i in range(len(start))])
 
-    pos = car_positions(start)
-    for car, place in enumerate(pos):
-        if final[place] == car:
+    # speed optimization, check lengths of sets are the same before checking sets
+    if len(start_set) != len(car_set) or len(final_set) != len(car_set):
+        return False
+
+    # elements of both the start and final sets should be 0 to len(start) - 1
+    return start_set == car_set and final_set == car_set
+
+
+def _invert(state: List[int]) -> List[int]:
+    """ Method to get the inverse of a list, i.e. value becomes the position and position becomes the value
+    Args:
+        state: a list of integers from 0 to len(state) - 1 e.g [1, 2, 0, 3]
+    Returns:
+        List(int): inverse of the list e.g [2, 0, 1, 3]
+    """
+    new_state = [0] * len(state)
+    for i, j in enumerate(state):
+        new_state[j] = i
+    return new_state
+
+
+def _move(car_positions: List[int], car_num: int) -> int:
+    """ Method to move car to current empty position
+    Args:
+        car_positions: list of integers where ith element is the current position of ith car
+        car_num: int number of car to move
+    Returns:
+        int: new empty position
+    """
+    car_positions[0], car_positions[car_num] = car_positions[car_num], car_positions[0]
+    return car_positions[0]
+
+
+def generate_moves(start: List[int], final: List[int]) -> List[int]:
+    """ Method to yield optimal sequence of moves in format (car number, old space, new space)
+    Args:
+        start: list of integers where ith element is the car currently in the ith space
+        final: list of integers where ith element is the car that should be in the ith space
+    Returns:
+        List[int]: returns final list of positions (this should be equal to final and is only for testing)
+    """
+
+    # inverted list of start list, i.e. ith element is current position of ith car
+    car_positions = _invert(start)
+    empty_place = car_positions[0]
+    for car, place in enumerate(car_positions):
+        # if car is already where it is supposed to be, do nothing
+        if car == final[place]:
             continue
 
-        if car:  # this is a cycle fix
-            yield car, place, pos[0]
-            pos[0], pos[car] = pos[car], pos[0]
+        # if car exists, break the cycle by moving car to empty position
+        if car:
+            yield car, place, empty_place
+            empty_place = _move(car_positions, car)
 
-        car = final[pos[0]]
-        while car:
-            yield car, pos[car], pos[0]
-            pos[0], pos[car] = pos[car], pos[0]
-            car = final[pos[0]]
+        # move car that should end up in current empty space while such car exists
+        while final[empty_place]:
+            current_car = final[empty_place]
+            yield current_car, car_positions[current_car], empty_place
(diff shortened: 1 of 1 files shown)


group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
+def car_positions(state):
+    """ given state returns array where ith element is the position of ith car """
+    positions = [0] * len(start)
```
SilverSolver: what does "start" mean? Did you mean "state"?

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
...
+    ith element of final state is the car that should be in ith space
+    empty space is 0
+    yields optimal sequence of moves in format (car number, old space, new space) """
+
+    pos = car_positions(start)
+    for car, place in enumerate(pos):
+        if final[place] == car:
+            continue
+
+        if car:  # this is a cycle fix
```
SilverSolver: I can't see, how car variable can become "None" or "False". So, for now, I don't know how this "if" statement can be not satisfied. Please, clarify, what are you planning to do to make this statement useful?

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
...
+            yield car, place, pos[0]
+            pos[0], pos[car] = pos[car], pos[0]
+
+        car = final[pos[0]]
+        while car:
```
SilverSolver: The same question here... I don't understand for now, if cars can be exhausted in some sense?

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
...
+            yield car, pos[car], pos[0]
```
SilverSolver: 1) So, if I understand correctly, for now your function just swaps each car with car at its final position until the positions of all the cars will be coincide with final. But I don't understand, why we need the recursion for this? I am confused. I'm not sure if it will work properly for now (I didn't test it yet).
2) Yes, swaping car1 and car2 can be considered as some simplification of 3 moves: car1 -> empty space; car2 -> car1's prev. position; car1 -> car2's prev. position. But it seems like we should yield all these 3 moves separately for assignment purpose.

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
...
+            pos[0], pos[car] = pos[car], pos[0]
+            car = final[pos[0]]
```
SilverSolver: I still don't understand the stopping conditions for this function clearly... are you sure that it will really stop every time?

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def fancy_output(start, final):
+    for i in generate_moves(start, final):
+        print("Move car number {} from {} to {}".format(*i))
+    print("Done")
+
+
+fancy_output([1, 0, 2, 3, 4], [2, 3, 1, 4, 0])
```
SilverSolver: I think, that these representations for start and final states are simple enough and are suitable for this assignment. I think that I will use them for mine solution, too, for being compatible with yours interfaces (I hope that tomorrow I will find a time for coding and submit something, too!).

group/assignment6/move_cars.py
crossopt: Oops, that was clumsy last-minute copy-paste. FIxed

It actually works now:)

group/assignment6/move_cars.py
crossopt: Possibly this should be "if car == 0" to make it more readable. However since 0 is an empty space, or basically a nonexistent car, I prefer this way. I.e. it can be read as "does this car actually exist"

group/assignment6/move_cars.py
crossopt: 2. I'm confused as to what you're saying. Every yield returns one move, I think

1. So, what's going on is I've already glued the empty space into the cycle and am now fixing the cycle. I'm moving the car that should end up in the empty space (final[pos[0]]) into the empty space (pos[0]) from its current position. Then I'm doing the same from the previous car in the cycle
I want to iterate over all the cars in the cycle and stop once they are all in the correct positions. This will be once the car that I am trying to move is 0 (there are two intuitive ways to understand. Firstly, this is the last car in the order of the cycle, since it was in the position I moved the car into. Secondly, the current car I'm looking at is always the car that should end up in the current empty position (final[pos[0]]). It's logical to stop when that car is the zero car)

group/assignment6/move_cars.py
crossopt: Long text here that probably should make its way into the code's comments sometime

1. The state representation I've chosen is this. I've numbered the cars from 1 to N-1 and the parking spaces from 0 to N, and decided to treat the empty parking space as having the car number 0. The advantage of this representation is that an array a[i] = car in ith parking space is a permutation (and b[i] = position of ith car is its inverse). Oh, and so are the movements that we want to do - a slightly simpler data structure might be to store one array c[i] = where should the car currently in c end up, and keep a variable that shows where the empty space currently is.
2. So, we know that we need to perform a specific permutation on the cars using only a series of allowed moves. Which are actually swaps of two elements, one of which is 0.
Since we know that every permutation is a set of disjoint cycles let's look at the solution for each cycle.
3. If a cycle of length K contains 0, we can perform it in K-1 moves by swapping 0 with each next element. Every move places one car in the place it should end up in the final state, so this is obviously optimal
If a cycle of length K doesn't contain 0, none of those cars can be moved to the place where they should end up in. Let's move one car to the current empty space. In my implementation the first cycle I fix is the cycle with 0, so by this point the empty space is the space that should be empty in the end. It's possible to check (a pen and a piece of paper are useful here) that now you have simply inserted 0 into the current cycle and so this can be solved like the previous case

group/assignment6/move_cars.py
SilverSolver: Thanks for clarifying! At first I didn't understand the meaning of the comment "empty space is 0". Now, if empty parking space is the car number 0, and that this "if" separates cases when car is 0 or not 0, everything fell into place.

group/assignment6/move_cars.py
SilverSolver: 2. I just didn't understand that zero car is the empty space. But I understand now.

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
...
+    return positions
+
+
+def generate_moves(start, final):
+    """ ith element of start state is the car currently in the ith space
+    ith element of final state is the car that should be in ith space
+    empty space is 0
+    yields optimal sequence of moves in format (car number, old space, new space) """
+
+    pos = car_positions(start)
```
Sridha2412: not sure why you would need car_positions(start). Isnt it essentially returning the same as start?

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def fancy_output(start, final):
+    for i in generate_moves(start, final):
+        print("Move car number {} from {} to {}".format(*i))
+    print("Done")
+
+
+fancy_output([1, 0, 2, 3, 4], [2, 3, 1, 4, 0])
+# fancy_output([1, 0, 2, 3, 4], [2, 0, 1, 4, 3])
```
Sridha2412: I tried printing the final pos and it didnt return the same as your output:
fancy_output([1, 0, 2, 3, 4], [2, 3, 1, 4, 0]) - [4, 2, 0, 1, 3]
fancy_output([1, 0, 2, 3, 4], [2, 0, 1, 4, 3]) - [1, 2, 0, 4, 3]

Sridha2412:
It took me a while to understand what you were doing, more comments would definitely be helpful :)

group/assignment6/move_cars.py
crossopt: car_positions returns the inverse of the permutation start

However start in my tests is its own inverse:)

group/assignment6/move_cars.py
crossopt: What exactly did you print?

I copied my code into an online python compiler and checked the output by hand. The sequence of moves outputted resulted in the end sequence

I'll double check when I get back to my computer, but I'm still curious as to what exactly you're printing

crossopt: Ooh, I know!

Sorry, my bad, my current on-computer version has different variable names:)

pos should be the inverse of the final state so those outputs are correct

crossopt: Because pos[i] is "where is the ith car now"
And final[i] is "which car should end up in the ith place"


I'll add some comments later today:)

Sridha2412:
Thanks for adding the comments and the great solution :D

SilverSolver (approved):
I agree to merge this version of solution and use it as a baseline for future improvements

PR #32, opened 2018-07-27
Silver solver 6

SilverSolver:
I used crossopt's solution as a baseline and added the validation of input + simplest tests

### DIFF
--- a/group/assignment6/move_cars.py
+++ b/group/assignment6/move_cars.py
@@ -1,7 +1,7 @@
 def car_positions(state):
     """ given state returns array where ith element is the position of ith car """
-    positions = [0] * len(start)
-    for place, car in enumerate(start):
+    positions = [0] * len(state)
+    for place, car in enumerate(state):
         positions[car] = place
     return positions
 
--- a/group/assignment6/move_cars.py
+++ b/group/assignment6/move_cars.py
@@ -1,38 +1,59 @@
-def car_positions(state):
-    """ given state returns array where ith element is the position of ith car """
-    positions = [0] * len(state)
-    for place, car in enumerate(state):
-        positions[car] = place
-    return positions
-
-
 def generate_moves(start, final):
-    """ ith element of start state is the car currently in the ith space
-    ith element of final state is the car that should be in ith space
+    """ ith element of start start is the car currently in the ith space
+    ith element of final start is the car that should be in ith space
     empty space is 0
     yields optimal sequence of moves in format (car number, old space, new space) """
 
-    pos = car_positions(start)
-    for car, place in enumerate(pos):
-        if final[place] == car:
-            continue
+    # validation of the input
+    start_set = set(start)
+    final_set = set(final)
+    #print(start_set, final_set)
+    if len(start_set) != len(start) or start_set != final_set or \
+       len(final_set) != len(final) or \
+       start_set != final_set and 0 not in start_set:
+        yield None
 
-        if car:  # this is a cycle fix
-            yield car, place, pos[0]
-            pos[0], pos[car] = pos[car], pos[0]
+    # pos will contain car positions in another form
+    pos = [0] * len(start)
+    if 0 in start_set:
+        for place, car in enumerate(start):
+            pos[car] = place
 
-        car = final[pos[0]]
-        while car:
-            yield car, pos[car], pos[0]
-            pos[0], pos[car] = pos[car], pos[0]
+        # main loop
+        for car, place in enumerate(pos):
+            if final[place] == car:
+                continue
+
+            if car:  # this is a cycle fix
+                yield car, place, pos[0]
+                pos[0], pos[car] = pos[car], pos[0]
+
             car = final[pos[0]]
-
+            while car:
+                yield car, pos[car], pos[0]
+                pos[0], pos[car] = pos[car], pos[0]
+                car = final[pos[0]]
 
 def fancy_output(start, final):
-    for i in generate_moves(start, final):
-        print("Move car number {} from {} to {}".format(*i))
+    result = generate_moves(start, final)
+    for i in result:
+        if i:
+            print("Move car number {} from {} to {}".format(*i))
+        else:
+            print("Unsolvable")
+            break
     print("Done")
 
+def return_output(start, final):
+    result = generate_moves(start, final)
+    output = []
+    for i in result:
+        if i:
+            output.append("Move car number {} from {} to {}".format(*i))
+        else:
+            output = ["Unsolvable"]
+            break
+    return output
 
-fancy_output([1, 0, 2, 3, 4], [2, 3, 1, 4, 0])
-# fancy_output([1, 0, 2, 3, 4], [2, 0, 1, 4, 3])
+if __name__ == "__main__":
+    fancy_output([1, 0, 2, 3, 4], [2, 3, 1, 4, 0])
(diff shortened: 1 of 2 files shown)


group/assignment6/move_cars.py
```diff
@@ -0,0 +1,59 @@
+def generate_moves(start, final):
+    """ ith element of start start is the car currently in the ith space
+    ith element of final start is the car that should be in ith space
+    empty space is 0
+    yields optimal sequence of moves in format (car number, old space, new space) """
+
+    # validation of the input
```
crossopt, line 7: This should probably be a separate function.
I think I remember something about each part of the code needing to have a separate responsibility?

It also seems logical to have validation be optional. And the validation for opt3 is going to be, mmm, less nontrivial, so makes sense to keep it separate from the beginning and have the opt3 validation reuse this piece

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,59 @@
...
+    start_set = set(start)
+    final_set = set(final)
+    #print(start_set, final_set)
+    if len(start_set) != len(start) or start_set != final_set or \
+       len(final_set) != len(final) or \
+       start_set != final_set and 0 not in start_set:
+        yield None
+
+    # pos will contain car positions in another form
+    pos = [0] * len(start)
```
crossopt, line 17: Same thing here about code structure. Should this be a logical part of the generation?

My personal preference is to have it in the same function, but, well
Competitive programming: if you can read it 5 minutes later, it's good code

I'm probably not the best judge here:)

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,59 @@
...
+                pos[0], pos[car] = pos[car], pos[0]
+                car = final[pos[0]]
+
+def fancy_output(start, final):
+    result = generate_moves(start, final)
+    for i in result:
+        if i:
+            print("Move car number {} from {} to {}".format(*i))
+        else:
+            print("Unsolvable")
```
crossopt, line 43: Unsolvable?
More like "Wrong input format"

Opt3 will actually have unsolvable cases (and I'll try to get around to writing it in the next couple of days), so best to keep the distinction

Maybe we want exceptions for wrong input format? Possibly checking the Google code style beforehand

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,59 @@
...
+            break
+    print("Done")
+
+def return_output(start, final):
```
crossopt, line 47: Why does this function exist?

I added an output function because looking through tuples got annoying very fast. However they are a simpler function and as such it makes more sense to return them anywhere the output of the generator is needed

group/assignment6/tests.py
```diff
@@ -0,0 +1,17 @@
...
+class TestMoveCars(unittest.TestCase):
+
+    def test_0(self):
+        self.assertEqual(return_output([1], [0]), ["Unsolvable"])
+        self.assertEqual(return_output([1, 2, 0], [1, 3, 0]), ["Unsolvable"])
+        self.assertEqual(return_output([1], [1]), [])
+        self.assertEqual(return_output([1, 2, 0], [1, 2, 0]), [])
+
+    def test_1(self):
+        self.assertEqual(return_output([1, 0], [0, 1]), ['Move car number 1 from 0 to 1'])
```
crossopt, line 13: Yeah, this is what I'm talking about. Why not check the result of the generator instead of weird strings?

group/assignment6/tests.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+from move_cars import *
+
+class TestMoveCars(unittest.TestCase):
+
+    def test_0(self):
```
crossopt, line 6: Ooh, neat corner cases!

The names and split of functions are confusing though.
If test_0 is supposed to be test_incorrect, why are the solvable cases there?
Maybe move them to test_1, add a few more wrong input cases (empty arrays, different array sizes, several of the same cars)

Or just rename the test functions. I might have misunderstood you completely:)

group/assignment6/move_cars.py
SilverSolver: I merged "car_positions" with "generate_moves" just because "car_positions" is called only once from "generate_moves".  So this merging makes the code shorter. I made validation the part of "generate_moves" for the same reason.
But I do not insist, because, yes, more functions can be more readable and testable, from the other side.
Let's see what others will say about separating code into logical blocks. I think that different approaches can exist there.

group/assignment6/move_cars.py
SilverSolver: Yes, I think that checking Google code style is a good idea. I will do it and write a conclusion about suitable treatment of a wrong input etc in comments.
I also think that we should discuss this moment with other group members, as a part of program design.

group/assignment6/move_cars.py
SilverSolver: Is is just a draft function which was written quickly and dirty and makes testing possible in some way. I agree that it should be redone later.

group/assignment6/tests.py
SilverSolver: You are right, I'll reorganize test functions later.

group/assignment6/move_cars.py
Sridha2412: having a different validation function could be a good idea. It will make it easy to read and test 👍

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,59 @@
...
+                continue
+
+            if car:  # this is a cycle fix
+                yield car, place, pos[0]
+                pos[0], pos[car] = pos[car], pos[0]
+
+            car = final[pos[0]]
+            while car:
+                yield car, pos[car], pos[0]
+                pos[0], pos[car] = pos[car], pos[0]
```
Sridha2412, line 34: we cant swap two cars, it will have to be three moves

group/assignment6/move_cars.py
```diff
@@ -0,0 +1,59 @@
...
+                car = final[pos[0]]
+
```
Sridha2412, line 36: it doesn't return the correct final position

group/assignment6/tests.py
```diff
@@ -0,0 +1,17 @@
...
+        self.assertEqual(return_output([1], [0]), ["Unsolvable"])
+        self.assertEqual(return_output([1, 2, 0], [1, 3, 0]), ["Unsolvable"])
+        self.assertEqual(return_output([1], [1]), [])
+        self.assertEqual(return_output([1, 2, 0], [1, 2, 0]), [])
+
+    def test_1(self):
+        self.assertEqual(return_output([1, 0], [0, 1]), ['Move car number 1 from 0 to 1'])
+        self.assertEqual(return_output([1, 2, 0], [1, 0, 2]), ['Move car number 2 from 1 to 2'])
+
```
Sridha2412, line 15: Good tests! Might want to rename the function names thought 😅

group/assignment6/move_cars.py
SilverSolver: pos[0] is not exactly the car, it is the empty space. It was discussed at the comments for crossopt's pull request for the same problem.

group/assignment6/move_cars.py
SilverSolver: Can you give an example of the input or write the test which will show it?

SilverSolver:
I'll better work in another branch from now on

PR #33, opened 2018-07-29
partial solution to question

Sridha2412:


Sridha2412:
@SilverSolver @crossopt I misinterpreted your code, it does return the right list and both your approaches might be more efficient but we might be better off using better function/variable names :)

PR #34, opened 2018-07-29
partial solution to assignment 6

Sridha2412:
@SilverSolver @crossopt  I misinterpreted your solution. Your approach is better than mine, I think but we should definitely use clearer function/variable names

### DIFF
--- a/group/assignment6/sridha2412.py
+++ b/group/assignment6/sridha2412.py
@@ -1,0 +1,80 @@
+from typing import List
+
+def _get_inverse(state: List[int]) -> List[int]:
+    """ Method to get the inverse of a list i.e value becomes the position and position becomes the value
+    Args:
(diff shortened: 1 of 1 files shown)


group/assignment6/sridha2412.py
```diff
@@ -0,0 +1,80 @@
...
+
+             # move current car to the now empty actual place
+            yield car, place, actual_place
+            start, start_inv = _swap(start, start_inv, actual_place, place)
+
+    # print(start)
+    return start
+
+
+def fancy_output(start, final):
```
Sridha2412, line 74: thanks @crossopt !

crossopt: I glued your comments onto my solution and committed it

Should be more readable now:)

PR #35, opened 2018-07-31
6th assignment

randomUsernameAndPassword:
Partial solution (without #3 and #4 optional problems) to assignment 6.

### DIFF
--- a/group/assignment6/rearrange_cars.py
+++ b/group/assignment6/rearrange_cars.py
@@ -1,0 +1,64 @@
+import unittest
+
+def inverse_map(state):
+    """ Returns array indexed with previous values. """
+    inversed = len(state) * [0]
+
+    for position, car in enumerate(state):
+        inversed[car] = position
+
+    return inversed
+
+def move_car(start, start_position, current_empty):
+    """
+        Moves car from start_position to empty place.
+        Returns new empty place and configuration with moved car.
+    """
+    start[start_position], start[current_empty], current_empty = start[current_empty], start[start_position], start_position
+    return (current_empty, start)
+
+def find_sequence(start, end):
+    """
+        Yields sequence of moves between start and end states.
+    """
+    current_empty = start.index(0)
+    end_positions = inverse_map(end)
+    
+    for position in range(len(start)):
+        # At first, move car, that should be placed in current
+        # empty space and repeat.
+        while end_positions[0] != current_empty:
+            start_positions = inverse_map(start)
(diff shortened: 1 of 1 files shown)


group/assignment6/rearrange_cars.py
```diff
@@ -0,0 +1,64 @@
...
+            # Move car from end position to empty place
+            if end_position != current_empty:
+                current_empty, start = move_car(start, end_position, current_empty)
+                yield start
+
+            # Move current car to end position
+            current_empty, start = move_car(start, position, current_empty)
+            yield start
+            
+def print_sequence(start, end):
```
Sridha2412, line 49: we should use crossopts function for this

group/assignment6/rearrange_cars.py
```diff
@@ -0,0 +1,64 @@
...
+
+def find_sequence(start, end):
+    """
+        Yields sequence of moves between start and end states.
+    """
+    current_empty = start.index(0)
+    end_positions = inverse_map(end)
+    
+    for position in range(len(start)):
+        # At first, move car, that should be placed in current
```
crossopt, line 28: Thanks for the coherent comments, they really made this function easy to understand:)

group/assignment6/rearrange_cars.py
```diff
@@ -0,0 +1,64 @@
...
+        # empty space and repeat.
+        while end_positions[0] != current_empty:
+            start_positions = inverse_map(start)
```
crossopt, line 31: Doing this in a cycle N times is way too slow

group/assignment6/rearrange_cars.py
crossopt: This name is far more reasonable though

group/assignment6/rearrange_cars.py
```diff
@@ -0,0 +1,64 @@
...
+            
+def print_sequence(start, end):
+    print(start)
+    for i in find_sequence(start, end):
+        print(i)
+    print("\n")
+
+class TestCarSequences(unittest.TestCase):
+    def test(self):
+        print_sequence([1, 2, 0, 3], [3, 1, 2, 0])
```
crossopt, line 57: Shouldn't the tests do more than print the output?

That said, it would be awesome if you could add some checks/asserts and merge Laida's test_0 onto them

SilverSolver: Yeah, I think we'd better use something like "yield_sequence" and then compare it's output with correct sequences somehow

PR #36, opened 2018-08-03
Some improvements and additions

SilverSolver:
Made 80 or less characters at each line of code in move_cars (according to Google code style). Also removed 1 mistake in move_cars. Added tests

### DIFF
--- a/group/assignment6/move_cars.py
+++ b/group/assignment6/move_cars.py
@@ -4,8 +4,10 @@ from typing import List
 def check_input(start: List[int], final: List[int]) -> bool:
     """ Method to check correctness of input
     Args:
-        start: list of integers where ith element is the car currently in the ith space
-        final: list of integers where ith element is the car that should be in the ith space
+        start: list of integers where ith element is the car currently in the
+               ith space
+        final: list of integers where ith element is the car that should be in
+               the ith space
     Returns:
         bool: is the input correct
     """
@@ -13,8 +15,11 @@ def check_input(start: List[int], final: List[int]) -> bool:
     final_set = set(final)
     car_set = set([i for i in range(len(start))])
 
-    # speed optimization, check lengths of sets are the same before checking sets
-    if len(start_set) != len(car_set) or len(final_set) != len(car_set):
+    # let's return False in the case of an empty sets
+    if len(start_set) == 0 or len(final_set) == 0:
+        return False
+
+    if len(start) != len(car_set) or len(final) != len(car_set):
         return False
 
     # elements of both the start and final sets should be 0 to len(start) - 1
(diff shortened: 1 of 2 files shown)


group/assignment6/move_cars.py
```diff
@@ -68,7 +80,8 @@ def generate_moves(start: List[int], final: List[int]) -> List[int]:
             yield car, place, empty_place
             empty_place = _move(car_positions, car)
 
-        # move car that should end up in current empty space while such car exists
+        # move car that should end up in current empty space while such a car
+        # exists
```
crossopt: This break looks a bit weird with one word on a new line

group/assignment6/move_cars.py
```diff
@@ -4,25 +4,31 @@
...
     Returns:
         bool: is the input correct
     """
     start_set = set(start)
     final_set = set(final)
     car_set = set([i for i in range(len(start))])
 
-    # speed optimization, check lengths of sets are the same before checking sets
-    if len(start_set) != len(car_set) or len(final_set) != len(car_set):
+    # let's return False in the case of an empty sets
```
crossopt: 'an empty set', probably

group/assignment6/tests.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertEqual(list(generate_moves([0], [0])), [])
+        self.assertEqual(list(generate_moves([1, 2, 0], [1, 2, 0])), [])
+
+    def test_1_move(self):
+        self.assertEqual(list(generate_moves([1, 0], [0, 1])), \
+			[(1, 0, 1)])
+        self.assertEqual(list(generate_moves([1, 2, 0], [1, 0, 2])), \
+			[(2, 1, 2)])
+
+    def test_correctness(self):
```
crossopt, line 52: So we check that the algorithm does less than two moves correctly manually, but if it's more we don't check the amount, only the correctness?

Strange but OK

group/assignment6/tests.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import List, Tuple
+from move_cars import *
+
+def check_movement_correctness(start: List[int], final: List[int], \
+                               moves: List[Tuple[int]]) -> bool:
+    # each move has format (car number, old space, new space)
+    current_pos = start[:]
+    for move in moves:
+        car, old_space, new_space = move
+        current_pos[old_space] = 0
```
crossopt: I think we want to check that the new space had a zero here
And that old_space had car
And swap

group/assignment6/move_cars.py
SilverSolver: Yep, I just don't know how to break it better if we should keep each line <= 80 characters

group/assignment6/tests.py
SilverSolver: Yes, I just didn't write the tests yet for checking more than two moves manually. It is a bit harder, because there are different correct possibilities. I would appreciate help with it!

randomUsernameAndPassword: As you noticed, there are different possibilities, so I would compare only lengths of lists produced by generate_moves.
I don't know if there is any reasonable way to prove, that given sequence is the shortest possible one, so I would just provide these lengths manually.

SilverSolver: Added additional test for lengths. Squashed previous commit into new (rebase was made).

Sridha2412:
good code and tests! :D I will give the other optional challenges a go.

crossopt (approved):

PR #37, opened 2018-08-07
4th followup

evinpinar:
This one searches for all possible paths which do not repeat the same configurations.


SilverSolver (commented):
Hi! That's cool that you made it. But it has another interface then merged part of our solution. We used lists for data representation, but you are using strings. So they are incompatible. I think that this is obstacle for merging :frowning_face:

